PR #1, opened 2015-08-31
Fixed area values in test

LuvCurves:


LuvCurves:
Using the updated module on the Puppet Master resolved the issue.

PR #13, opened 2015-10-05
Added 'route-map' support to bgp_neighbor_af

chrisvanheuveln:
route-map in/out have the same behaviors as filter-list in/out so this is just a copy/replace for the most part.
- rubocop'd
- beaker'd (tests added & run)
- Docs updated

### DIFF
--- a/lib/puppet/type/cisco_bgp_neighbor_af.rb
+++ b/lib/puppet/type/cisco_bgp_neighbor_af.rb
@@ -59,6 +59,8 @@ Puppet::Type.newtype(:cisco_bgp_neighbor_af) do
       max_prefix_interval                    => 30,
       next_hop_self                          => 'true',
       next_hop_third_party                   => false,
+      route_map_in                           => 'rm_in',
+      route_map_out                          => 'rm_out',
       route_reflector_client                 => true,
       send_community                         => 'extended',
       soft_reconfiguration_in                => true,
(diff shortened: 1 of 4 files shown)


lib/puppet/type/cisco_bgp_neighbor_af.rb
```diff
@@ -387,6 +389,24 @@ def self.title_patterns
     newvalues(:true, :false, :default)
   end
 
+  newproperty(:route_map_in) do
+    desc 'route-map in state. Valid values are a string defining the name ' \
```
mikewiebe: We used to provide more error checking... for instance, won't this error with a cryptic message if the user passes in an integer or fixnum instead of a string?

chrisvanheuveln: How about just munging it to_s?

   munge do |value|
     value = :default if value == 'default'
     value.to_s
   end

mikewiebe: The problem is that we still don't do early detection and failure with this change.  This would only fail during the config_set when the cli rejects an integer as an invalid route-map name.  We need to explicitly fail using the 'validate' method in the type to give clear user guidance.

validate do |value|
      fail("Property 'route_map_in' must be a string") unless
        value.is_a? String
    end

    munge do |value|
      value = :default if value == 'default'
      value
    end

chrisvanheuveln: I think you miss my point. Typecasting to string is a good thing, but an integer can be a valid route-map name so the cli won't reject it. I don't think the string check is necessary - can you give me an example where to_s does not solve this problem?

mikewiebe: No. I did not miss your point, but I did incorrectly assume that route-map could *NOT* be an integer :).  Looks good.

:+1:

PR #14, opened 2015-10-07
(WIP) implementing netdev_stdlib network_vlan

tphoney:
This is work in progress. It does however work.

### DIFF
--- a/lib/puppet/provider/network_vlan/nxapi.rb
+++ b/lib/puppet/provider/network_vlan/nxapi.rb
@@ -1,0 +1,57 @@
+$LOAD_PATH.unshift(File.join(File.dirname(__FILE__),"..","..",".."))
+require 'puppet/type/cisco_vlan'
+require 'puppet/provider/cisco_vlan/nxapi'
+
+Puppet::Type.type(:network_vlan).provide(:nxapi, :parent => Puppet::Type.type(:cisco_vlan).provider(:nxapi)) do
+  @doc = "cisco VLAN"
+
+  has_features :activable, :describable
+
+  mk_resource_methods
+
+  def self.instances
+    vlans = []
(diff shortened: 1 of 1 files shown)


tphoney:
@chrisvanheuveln 
Agreed WIP code is not to be merged, nor do i want it to.

this is a stdlib_netdev provider, this allows puppet users to have a standard way of creating network resources across multiple network device providers. ie 
the same manifest will create a vlan on an F5, cisco, junos, mlnx-os, arista ....

The existing cisco_vlan provider is specific to cisco_vlan type. ie not stdlib_netdev network_vlan type

glennmatthews:
Chris's point (which I agree to) is why are you reimplementing the flush and instances methods - are the implementations inherited from cisco_vlan insufficient?

tphoney:
They are both class methods that are not inherited.
and
The mappings for the properties / parameters are not 1-1.

netdev-stdlib             | cisco-network-puppet-module provider
description                | (not in cisco)
id                               | vlan	 
shutdown	                  | shutdown	 
vlan_name	          | vlan_name	 
(not in stdlib_netdev) | state	 

Because there are parameters not used we cannot use your flush / self.instances

jonnytpuppet:
@chrisvanheuveln @glennmatthews Typically, regarding Puppet modules design patterns, there is a 1-many relationship between types and providers (i.e. one type can have multiple providers). However, the reverse case is not true. As such, a provider may only be restricted to one type.

@tphoney is calling methods inside your provider to aid future changes, where if something were to change to these properties, you'd only have to modify them once.

The alternative, in my eyes, is that we duplicate the code that is already in your getter/setter methods, and put it inside this new provider.

Cheers

JT

jonnytpuppet:
@tphoney looking at your code again, it seems like you're calling directly in to the util gem anyway? Isn't Cisco::Vlan the util gem class?

PR #15, opened 2015-10-08
Added NXOS provider for network_snmp

jonnytpuppet:
tests/beaker_tests/network_snmp/network_snmp-provider-defaults.rb passed in 17.47 seconds
      Test Suite: tests @ 2015-10-08 13:57:25 +0100

      - Host Configuration Summary -

              - Test Case Summary for suite 'tests' -
       Total Suite Time: 17.46 seconds
      Average Test Time: 17.46 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1

PR #16, opened 2015-10-08
Apply Rubocop to puppet-module

robert-w-gries:
This is a very large set of changes.  If you have any suggestions for making code review easier, please let me know and I'll submit a different PR.

Testing:  All beaker tests passed except for fileservicepkg.

image


robert-w-gries:
Travis error is due to this:  https://gist.github.com/renier/9d23ead9a36ba2939d18

robert-w-gries:
Travis pulled in the additions from [this pull request](https://github.com/cisco/cisco-network-puppet-module/pull/15), which makes rubocop fail.  Everything else passes.

lib/puppet/type/cisco_interface_ospf.rb
```diff
@@ -240,34 +246,33 @@ def name
...
-          self[:message_digest_password].nil?)
+    if (!self[:message_digest_key_id].nil?) &&
+       (self[:message_digest_algorithm_type].nil? ||
+        self[:message_digest_encryption_type].nil? ||
+        self[:message_digest_password].nil?)
       fail("If message_digest_key_id is present in the manifest, \
-	    the following attributes must also be present - \
+      the following attributes must also be present - \
             message_digest_algorithm_type, \
             message_digest_encryption_type, message_digest_password")
```
glennmatthews: Only loosely related to your change, but this message and the one above it seem wrong to me - the backslashes suggest that this is meant to be a multi-line wrap but the lines aren't quoted - shouldn't it be:

```ruby
       fail('If message_digest_key_id is present in the manifest, ' \
            'the following attributes must also be present - ' \
            'message_digest_algorithm_type, ' \
            'message_digest_encryption_type, message_digest_password')
```

lib/puppet/type/cisco_interface_ospf.rb
```diff
@@ -240,34 +246,33 @@ def name
...
     end
-    
-    if self[:passive_interface] and
-      !/^lo\S+$/.match(self[:interface].downcase).nil?
-      fail "passive_interface value cannot be set on loopback interfaces"
-    end
 
+    if self[:passive_interface] &&
+       !/^lo\S+$/.match(self[:interface].downcase).nil?
```
glennmatthews: Not specific to your change but `!regex.match(str).nil?` is equivalent to simply `regex.match(str)`

lib/puppet/type/cisco_snmp_group.rb
```diff
@@ -51,18 +51,17 @@ def self.title_patterns
...
+      ],
     ]
-    return patterns
+    patterns
   end
 
   # Overwrites default name method.
   def name
-    return "#{self[:group]}"
+    "#{self[:group]}"
```
glennmatthews: String interpolation is not necessary - this could be just `self[:group]`

tests/beaker_tests/accessvlan/accessvlan_provider_defaults.rb
```diff
@@ -45,118 +45,117 @@
...
+                                          { 'ensure'                       => 'present',
+                                            'access_vlan'                  => '128',
+                                            'ipv4_proxy_arp'               => 'false',
+                                            'ipv4_redirects'               => 'true',
+                                            'negotiate_auto'               => 'true',
+                                            'shutdown'                     => 'false',
+                                            'switchport_autostate_exclude' => 'false',
+                                            'switchport_mode'              => 'access',
+                                            'switchport_vtp'               => 'false' },
+                                          false, self, logger)
```
glennmatthews: I think I commented on this on the earlier diffs - it would probably be more readable to instead do:

```ruby
       UtilityLib.search_pattern_in_output(
         stdout,
         {
           'ensure'                      => 'present',
           'access_vlan'                  => '128',
           'ipv4_proxy_arp'               => 'false',
           'ipv4_redirects'               => 'true',
           'negotiate_auto'               => 'true',
           'shutdown'                     => 'false',
           'switchport_autostate_exclude' => 'false',
           'switchport_mode'              => 'access',
           'switchport_vtp'               => 'false',
         }, false, self, logger)
```

Similar comment throughout - however I know we're planning to refactor a lot of this code in the future so I don't think this is necessarily something that needs to be fixed right now.

glennmatthews:
Looks sane to me. I have a few nitpicks but I'd rather get this merged in its current state and do additional post-fixup rather than hold off this whole PR any longer.

:+1:

lib/puppet/type/cisco_interface_ospf.rb
robert-w-gries: Both methods are valid, but the current implementation **is** incorrect.

    irb

    fail("If message_digest_key_id is present in the manifest, \
       the following attributes must also be present - \
             message_digest_algorithm_type, \
             message_digest_encryption_type, message_digest_password")

    RuntimeError: If message_digest_key_id is present in the manifest,       the following attributes must also be present -              message_digest_algorithm_type,              message_digest_encryption_type, message_digest_password
	from (irb):9
	from /usr/local/rvm/rubies/ruby-1.9.3-p551/bin/irb:12:in `<main>'

The indentation of the escaped new lines are being counted as spaces in the string.  Your method will work and is cleaner than the current implementation, so I agree we should change it.

robert-w-gries:
I'm going to merge the PR now, and apply Rubocop to any recently merged PRs. We should address your comments in the next refactoring PR.

PR #21, opened 2015-10-16
Add network_dns provider for netdev_stdlib

hunner:
This adds the network_dns provider, which uses the node-utils apis for
DomainName, DnsDomain, and NameServer for it's various properties.

### DIFF
--- a/tests/beaker_tests/netdev_stdlib/network_dns_provider_nondefaults.rb
+++ b/tests/beaker_tests/netdev_stdlib/network_dns_provider_nondefaults.rb
@@ -1,0 +1,252 @@
+###############################################################################
+# Copyright (c) 2014-2015 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


tests/beaker_tests/netdev_stdlib/network_dns_provider_nondefaults.rb
```diff
@@ -0,0 +1,252 @@
...
+    # Expected exit_code is 0 since this is a puppet resource cmd.
+    # Flag is set to false to check for presence of RegExp pattern in stdout.
+    cmd_str = UtilityLib.get_namespace_cmd(agent, UtilityLib::PUPPET_BINPATH +
+      "resource network_dns 'settings'", options)
+    on(agent, cmd_str) do
+      UtilityLib.search_pattern_in_output(stdout,
+                                          { 'ensure'  => 'present',
+                                            'domain'  => 'switch1.test.com',
+                                            'search'  => "\\['test.com', 'test.net'\\]",
+                                            'servers' => "\\['8.8.8.8', '2001:4860:4860::8888'\\]" },
```
glennmatthews: Why are the backslash escapes needed?

hunner: Apparently to match arrays of values from `puppet resource` it is a literal string of `[stuff]`, but it is evaluated as a regex character group. Escaping the `\[` once loses the \'s, so double escape is needed to get a literal `[` passed to the output matcher.

glennmatthews: Hmm, OK. Probably something that needs to be changed in our library code - I'll take a note to look into this as time permits, but fine for now.

glennmatthews:
:+1:

PR #22, opened 2015-10-16
Add prefix_list property to bgp_neighbor_af

chrisvanheuveln:
This is a boilerplate prop modeled on filter-list and route-map.

rubocop/beaker tests are clean.


chrisvanheuveln:
This PR should wait for my node_utils PR to be merged first.

lib/puppet/type/cisco_bgp_neighbor_af.rb
```diff
@@ -384,6 +386,24 @@ def self.title_patterns
     newvalues(:true, :false, :default)
   end
 
+  newproperty(:prefix_list_in) do
+    desc 'prefix-list in state. Valid values are a string defining the name ' \
+         "of the prefix-list or 'default'."
+    munge do |value|
+      value = :default if value == 'default'
```
hunner: Given that the next line casts `value` to a string, this line isn't needed is it?

chrisvanheuveln: It is needed -- it's the value.to_s that's wrong and that's breaking our default handling. Unfortunately we used this same code in a couple of places so I've got more to fix than just this one. New diff coming up...

mikewiebe:
:+1:

PR #30, opened 2015-10-22
add maxas-limit to bgp

andy91315:
1. pass the beaker test cd 
2. pass static analysis of type and provider

maxas-limit range 1~2000
change default value to 0 in node_util


andy91315:
change value range 1 to 512 check on N9k

chrisvanheuveln:
:+1:

PR #33, opened 2015-10-26
Added dummy NXOS provider for radius

jonnytpuppet:
This is a dummy provider. That is, when a manifest has enable set to either true or false, the provider will come back as if the same value is already set.

### DIFF
--- a/tests/beaker_tests/radius/radius_provider_defaults.rb
+++ b/tests/beaker_tests/radius/radius_provider_defaults.rb
@@ -1,0 +1,101 @@
+###############################################################################
+# Copyright (c) 2014-2015 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


tests/beaker_tests/radius/radiuslib.rb
```diff
@@ -0,0 +1,70 @@
...
+# A. Methods to create manifests for radius Puppet provider test cases.
+###############################################################################
+
+# Require UtilityLib.rb path.
+require File.expand_path('../../lib/utilitylib.rb', __FILE__)
+
+# A library to assist testing radius resource
+module RadiusLib
+  # Group of Constants used in negative tests for radius provider.
+  ENSURE_NEGATIVE = 'unknown'
```
glennmatthews: unused?

tests/beaker_tests/radius/radius_provider_defaults.rb
```diff
@@ -0,0 +1,101 @@
...
+
+    logger.info('Setup switch for provider')
+  end
+
+  # @step [Step] Requests manifest from the master server to the agent.
+  step 'TestStep :: Get resource present manifest from master' do
+    # Expected exit_code is 0 since this is a bash shell cmd.
+    on(master, RadiusLib.create_radius_manifest_true)
+
+    # Expected exit_code is 2 since this is a puppet agent cmd with change.
```
glennmatthews: Comment doesn't match actual code below (which expects 0, not 2). Same issue on next test step too.

jonnytpuppet:
@glennmatthews fixed, thanks for the quick feedback! :)

PR #39, opened 2015-10-29
Alpha-order bgp_af properties

chrisvanheuveln:
It's not obvious from the diffs but there are no functional changes in this commit; I'm simply moving
the references into alpha order.

rubocop/beaker are clean


mikewiebe:
:+1:

PR #51, opened 2015-11-03
Added NXOS provider for tacacs_global

jonnytpuppet:


jonnytpuppet:
@mikewiebe @glennmatthews Can this be merged please? rubocop is not failing on my local system, even after rebasing with develop, so I don't know why it's failing here.

jonnytpuppet:
@mikewiebe @glennmatthews travisci is checking again cisco's own code, which is failing.

mikewiebe:
@jonnytpuppet This is failing because of a newer version of rubcop that travis CI is using.  I have a fix for this coming soon.

mikewiebe:
@jonnytpuppet Can you provide the minitest results, update the demo manifest and README doc.

lib/puppet/provider/tacacs_global/nxapi.rb
```diff
@@ -0,0 +1,136 @@
...
+    v = Cisco::TacacsServer.new(false)
+
+    current_state = {}
+    current_state[:name] = name
+    current_state[:enable] = Cisco::TacacsServer.enabled ? :true : :false
+
+    if current_state[:enable].eql?(:true)
+      current_state[:timeout] = v.timeout
+      current_state[:key] = v.encryption_password
+      current_state[:key_format] = v.encryption_type
```
hunsbea: the value returned here is something like Cisco::TACACS_SERVER_ENC_CISCO_TYPE_7, I don't see where this is getting munged into a puppet-like value

lib/puppet/provider/tacacs_global/nxapi.rb
```diff
@@ -0,0 +1,136 @@
...
+      elsif @resource[:enable].eql?(:false)
+        @tacacs_global.destroy
+      end
+    end
+
+    # Handle timeout setting
+    @tacacs_global.timeout = @resource[:timeout] if @resource[:timeout]
+
+    # Handle key and keyformat setting
+    @tacacs_global.send('encryption_key_set', munge_flush(@resource[:key_format]), @resource[:key]) if @resource[:key]
```
hunsbea: Similar to above, this method expects a key format like Cisco::TACACS_SERVER_ENC_CISCO_TYPE_7 or other constant, but the resource value should be something like '7'? I don't see how this can work

lib/puppet/provider/tacacs_global/nxapi.rb
```diff
@@ -0,0 +1,136 @@
...
+    if @resource[:enable]
+      if @resource[:enable].eql?(:true)
+        @tacacs_global.enable
+      elsif @resource[:enable].eql?(:false)
+        @tacacs_global.destroy
+      end
+    end
+
+    # Handle timeout setting
+    @tacacs_global.timeout = @resource[:timeout] if @resource[:timeout]
```
hunsbea: looks like timeout can get set every time and won't be idempotent

lib/puppet/provider/tacacs_global/nxapi.rb
```diff
@@ -0,0 +1,136 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'cisco_node_utils' if Puppet.features.cisco_node_utils?
+begin
+  require 'puppet_x/cisco/autogen'
```
hunsbea: doesn't look like autogen is getting used anywhere, and require can be removed

lib/puppet/provider/tacacs_global/nxapi.rb
jonnytpuppet: This is an Enum, so it's just a number. This has been tested and worked 100%

lib/puppet/provider/tacacs_global/nxapi.rb
jonnytpuppet: Same as above, this is an Enum, so it's just a number. This has been tested and worked 100%

lib/puppet/provider/tacacs_global/nxapi.rb
jonnytpuppet: This is idempotent, as by resetting it to the same value won't do any harm. This is how we'd done all the other providers for you guys. This is how a flushing provider works (i.e. it sends everything).

tests/beaker_tests/tacacs_global/tacacs_globallib.rb
```diff
@@ -0,0 +1,89 @@
...
+# ---------------------
+# tacacs_globallib.rb
+#
+# This is the utility library for the Tacacs Global provider Beaker test cases
+# that contains the common methods used across the Radius Server testsuite's
+# cases. The library is implemented as a module with related methods and
+# constants defined inside it for use as a namespace. All of the methods are
+# defined as module methods.
+#
+# Every Beaker Radius Server test case that runs an instance of Beaker::TestCase
```
hunsbea: Radius

lib/puppet/provider/tacacs_global/nxapi.rb
hunsbea: Ah yeah I just refreshed and saw your comments, thanks

lib/puppet/provider/tacacs_global/nxapi.rb
mikewiebe: @jonnytpuppet We may have missed this in previous reviews.  We use flush in our other providers as well, but if the values are already set they don't get set again.

@hunsbea can you provide any thoughts on a fix for this?

hunsbea: should be a quick check to compare against current state before setting

jonnytpuppet: What is the benefit here of not setting everything when only some things have changed? If you want to check, you have to poll the device, which arguably could be more resource intensive.

A common pattern that is used in other puppet modules is that once one thing is changed, all properties are updated.

It is worth noting that if nothing has changed, then flush isn't called.

hunsbea: @jonnytpuppet 
But aren't you already retrieving the timeout here?
```
current_state[:timeout] = v.timeout
```
If my understanding of the puppet philosophy is correct, setting everything again if one thing has changed isn't a good idea - only properties that have changed should be updated, so that the resource can make progress toward the intended state.

And it looks to me like flush will always get called as long as the user specifies a password in the manifest, because passwords can't be compared directly?

jonnytpuppet: According to the Puppet Philosophy, a property such as password shouldn't be applicable to a device that doesn't return the same as it's input. I don't like the idea that it's there as all.

We'll need to go back to the drawing board and see if we can save the "original values" for access later. The snipped you referenced isn't accessable by the time flush is called.

jonnytpuppet: Looking at a lot of the other cisco-implemented providers, they also set everything.

@mikewiebe can you please provide an example of where only certain things are set?

hunsbea: @jonnytpuppet I agree idempotency doesn't need to apply to the password here, that's why I only made the idempotence comment on timeout. you don't need to do anything fancy to save the password or anything like that to try to make password idempotent, I was just mentioning this as a reason that flush will get called. And really? I thought that original value got saved into property_hash, flush can't access that?

hunsbea: An example where only certain things are set is the cisco global tacacs object: https://github.com/cisco/cisco-network-puppet-module/blob/develop/lib/puppet/provider/cisco_tacacs_server/nxapi.rb

There may be a simpler way to accomplish the same behavior, but the point is that only the password/encryption type are not idempotent

jonnytpuppet: @hunsbea it's important to point out that by simply flushing everything, doesn't make the system non-idempotent. The current implementation where everything is flushed is still idempotent.

jonnytpuppet: @hunsbea no, property_hash is not guaranteed to store the original values. Actually, the mk_resource_methods call will cause setter to be created at run-time that update property_hash with the new values.

What we can do here, is save the original values in the constructor, and compare during flush.

mikewiebe:
@jonnytpuppet , @hunsbea  and I chatted and agreed that the idempotence issue is not blocking for this PR.  Let's chat more about it at our next syncup.

jonnytpuppet:
tests/beaker_tests/tacacs_global/tacacs_global_provider_defaults.rb passed in 30.99 seconds
      Test Suite: tests @ 2015-11-12 14:55:16 +0000

jonnytpuppet:
@mikewiebe @glennmatthews Can this be merged ASAP please? Thanks

lib/puppet/provider/tacacs_global/nxapi.rb
mikewiebe: @jonnytpuppet Can you double check if autogen is needed?

jonnytpuppet:
@mikewiebe Autogen remove

mikewiebe:
@hunsbea  Please review latest diffs.

jonnytpuppet:
@mikewiebe @hunsbea @robert-w-gries Can this be merged now? Please ignore if the README is a little off, I'll fix it in another merge. It's getting difficult to fix the conflicts 100%.

jonnytpuppet:
@mikewiebe @hunsbea @robert-w-gries Merge conflicts fixed.

jonnytpuppet:
@mikewiebe @glennmatthews @hunsbea @robert-w-gries This still hasn't been merged yet. Please can this be merged? Thanks

PR #52, opened 2015-11-03
BGP Updates

mikewiebe:
This update addresses the following.

1) With our more complex title patterns, bgp resources were not idempotent.  This was fixed by overriding the name method to return the correct set of values matching the title.
2) Update to demo_bgp.pp to use complex title patterns for all examples and enable additional functionality.

### DIFF
--- a/examples/demo_bgp.pp
+++ b/examples/demo_bgp.pp
@@ -19,10 +19,9 @@ class ciscopuppet::demo_bgp {
   # --------------------------------------------------------------------------#
   # Configure Global BGP                                                      #
   # --------------------------------------------------------------------------#
-  cisco_bgp { 'default':
+  cisco_bgp { '55.77 blue':
     ensure                                 => present,
-    asn                                    => 55.77,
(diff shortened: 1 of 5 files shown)


examples/demo_bgp.pp
```diff
@@ -133,13 +128,66 @@
...
+  cisco_bgp_neighbor {'55.77 blue 2.2.2.2':
+    ensure                 => present,
+
+    #Properties
+    description            => 'my description',
+    connected_check        => true,
+    capability_negotiation => true,
+    dynamic_capability     => true,
+    ebgp_multihop          => 2,
+    local_as               => 55.77,
```
chrisvanheuveln: Non-blocker: local-as should be different than the global asn. It also should process _after_ remote-as or else it will throw an error. I think for now just remove local-as from here and create a US for forcing the processing order in the provider.

mikewiebe: Fixed

PR #66, opened 2015-11-10
modify aaa group tacacs to be compatible with the new netdev compatib…

hunsbea:
beaker passes, rubocop passes

I'll merge this at the same time I merge the node utils changes. This also addresses using autogen to gen a couple of the properties.

### DIFF
--- a/lib/puppet/provider/cisco_aaa_group_tacacs/nxapi.rb
+++ b/lib/puppet/provider/cisco_aaa_group_tacacs/nxapi.rb
@@ -18,6 +18,12 @@
 # limitations under the License.
 
 require 'cisco_node_utils' if Puppet.features.cisco_node_utils?
+begin
+  require 'puppet_x/cisco/autogen'
+rescue LoadError
+  require File.expand_path(File.join(File.dirname(__FILE__), '..', '..', '..',
+                                     'puppet_x', 'cisco', 'autogen.rb'))
+end
 
 Puppet::Type.type(:cisco_aaa_group_tacacs).provide(:nxapi) do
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/cisco_aaa_group_tacacs/nxapi.rb
```diff
@@ -101,59 +111,32 @@ def properties_set(new_aaa_group=false)
   end
   # rubocop:enable GuardClause
```
chrisvanheuveln: Still need this cop bypass?

lib/puppet/provider/cisco_aaa_group_tacacs/nxapi.rb
```diff
@@ -101,59 +111,32 @@ def properties_set(new_aaa_group=false)
...
-    @property_flush[:deadtime] = set_value
-  end
-
+  # can't autogen server_hosts, special array handling
   def server_hosts
     return [:default] if @resource[:server_hosts] &&
                          @resource[:server_hosts][0] == :default &&
                          @property_hash[:server_hosts] ==
-                         Cisco::AaaServerGroup.default_servers
+                         @aaa_group.default_servers
```
chrisvanheuveln: nit: this would be cleaner if you do:

```
    return [:default] if
      @resource[:server_hosts] &&
      @resource[:server_hosts][0] == :default &&
      @property_hash[:server_hosts] == @aaa_group.default_servers
```

your call

chrisvanheuveln: :+1: just nits

PR #68, opened 2015-11-12
Changes need in bgp provider to rescue unsupported features

RichWellum:
Adding rescues where needed.


lib/puppet/provider/cisco_bgp/nxapi.rb
```diff
@@ -230,7 +232,11 @@ def puts_config
     current = sprintf("\n%30s: %s", 'asn', @bgp_vrf.asnum)
     current.concat(sprintf("\n%30s: %s", 'vrf', @bgp_vrf.vrf))
     BGP_ALL_PROPS.each do |prop|
-      current.concat(sprintf("\n%30s: %s", prop, @bgp_vrf.send(prop)))
+      begin
+          current.concat(sprintf("\n%30s: %s", prop, @bgp_vrf.send(prop)))
+        rescue Cisco::UnsupportedError
+          nil
+        end
```
robert-w-gries: Indentation is off.  What's strange is that I ran rubocop 0.34.2 against this file, and it's not catching it.

lib/puppet/provider/cisco_bgp/nxapi.rb
```diff
@@ -230,7 +232,11 @@ def puts_config
...
     end
     debug current
   end # puts_config
```
chrisvanheuveln: :+1: after indentation fix

glennmatthews:
This is no longer needed after cisco/cisco-network-node-utils#82, correct?

RichWellum:
Yes correct - will have a new diff out shortly.

PR #71, opened 2015-11-16
Feature/aaa

hunsbea:
merge aaa group tacacs into develop now that the corresponding node utils piece has been merged and this provider/type can be used


chrisvanheuveln:
Alex said this is the same code that I just merged into feature/aaa so I'll merge after he posts test results.

hunsbea:
beaker tests ran successfully when I manually applied this hotfix: https://github.com/cisco/cisco-network-node-utils/commit/077408fd7b3be997ae6d6f2897e19136028e78bd which is already in develop, so this is all set.

PR #73, opened 2015-11-17
Feature/vxlan

alok-aggarwal:
Type and provider for vxlan_global.

Ran Rubocop and puppet-lint. Tested properties.


lib/puppet/type/cisco_vxlan_global.rb
```diff
@@ -0,0 +1,119 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+Puppet::Type.newtype(:cisco_vxlan_global) do
+  @doc = "Manages the vxlan global configuration of a Cisco device.
+
+  cisco_vxlan_global { <title>:
+    ..attributes..
+  }
+
```
chrisvanheuveln: Use markdown syntax to display these as "puppet" code, ie:

```
\~~~puppet
  cisco_vxlan_global { <title>:
    ..attributes..
  }
\~~~
```

lib/puppet/type/cisco_vxlan_global.rb
```diff
@@ -0,0 +1,119 @@
...
+  There can only be one instance of the cisco_vtp.
+  Example:
+    cisco_vxlan_global { default:
+      ensure   					=> present,
+      dup_host_ip_addr_detection_host_moves  	=> 200,
+      dup_host_ip_addr_detection_timeout	=> 20,
+      anycast_gateway_mac  			=> '1223.3445.5668',
+      dup_host_mac_detection_host_moves  	=> 100,
+      dup_host_mac_detection_timeout		=> 10,
+    }
```
chrisvanheuveln: MD fence blocks for this one too. See the bgp types for examples.

lib/puppet/type/cisco_vxlan_global.rb
```diff
@@ -0,0 +1,119 @@
...
+  newproperty(:dup_host_ip_addr_detection_timeout) do
+    desc "The duplicate detection timeout in seconds for the number of host moves.
+          Valid values are Integer, keyword 'default'."
+
+    munge do |value|
+      begin
+        value = :default if value == 'default'
+        value = Integer(value) unless value == :default
+      rescue
+        raise 'Dup IP Addr timeout must be an integer.'
```
chrisvanheuveln: It may be more helpful to just use the property names on these messages, ie:

```
        raise 'dup_host_ip_addr_detection_timeout must be an integer.'
```

lib/puppet/type/cisco_vxlan_global.rb
```diff
@@ -0,0 +1,119 @@
...
+      value
+    end
+  end # property dup IP addr timeout
+
+  newproperty(:anycast_gateway_mac) do
+    desc 'Distributed gateway virtual MAC address. Valid values are string.'
+
+    validate do |anycast_gateway_mac|
+      fail 'Mac Address is not a string.' unless anycast_gateway_mac.is_a? String
+    end
```
chrisvanheuveln: I did say that it doesn't make sense for some properties to support 'default'; however, in this case now I'm wondering about the scenario where there's an existing gateway mac in the configuration and the user wants to make sure it's not present.
What do you set in the manifest to remove it? For that reason I think we need support for default after all (for this property). Do you agree?

chrisvanheuveln:
Mostly just minor formatting comments and the outstanding question on default support. The rest is good.

lib/puppet/type/cisco_vxlan_global.rb
chrisvanheuveln: Please update CHANGELOG.md.

alok-aggarwal: If the user sets anycast_gateway_mac => ' ' in the manifest file, it will remove the existing mac address.

alok-aggarwal: I was not sure what exactly to add, so i added this. Is this okay?
## Unreleased
### Added
- New type and provider for cisco_vxlan_global which covers the following attributes:
  - feature fabric forwarding
  - fabric forwarding dup-host-ip-addr-detection <host_moves> <timeout>
  - l2rib dup-host-mac-detection <host_moves> <timeout>
  - fabric forwarding anycast-gateway-mac <mac-addr>

lib/puppet/type/cisco_vxlan_global.rb
alok-aggarwal: done

lib/puppet/type/cisco_vxlan_global.rb
alok-aggarwal: done

lib/puppet/type/cisco_vxlan_global.rb
alok-aggarwal: done

lib/puppet/type/cisco_vxlan_global.rb
chrisvanheuveln: CHANGELOG: Actually it just needs to be a one liner under "New Features" like: `cisco_vxlan_global type and provider`. You can leave it like this for now and I'll fix it post-merge.

Defaults: Yes, that does work. However, consider this from a user's perspective: you clean up all of the other properties using 'default' in the manifest, but you have to remember that this particular property doesn't follow that convention and that it instead needs ' '. That seems like an unnecessary restriction for the user.

alok-aggarwal: Okay, supporting 'default' for anycast_gateway_mac now.

PR #74, opened 2015-11-17
grpc_port parameter supported in hosts cfg file

robert-w-gries:
Adds support for grpc_port parameter in hosts.cfg file and allows connection to grpc port on switch.

### DIFF
--- a/docs/README-develop-beaker-scripts.md
+++ b/docs/README-develop-beaker-scripts.md
@@ -74,6 +74,8 @@ $ cd cisco-network-puppet-module/tests/beaker_tests/
 
 Under the `beaker_tests` directory, create file named `host.cfg` and add the following content.
 
+Note: If running puppet on XR, specify the gRPC port number configured on the switch.
+
 Replace the `< >` markers with specific information.
 
 ```bash
@@ -83,6 +85,7 @@ HOSTS:
             - agent
         platform: cisco-7-x86_64
         ip: <fully qualified domain name>
+        grpc_port <grpc port number for XR switch>
         vrf: <vrf used for beaker workstation and puppet master ip reachability>
         ssh:
           auth_methods: ["password"]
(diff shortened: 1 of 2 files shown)


glennmatthews:
Does this (especially the removal of 'sudo') break nexus?

robert-w-gries:
Yes it does. Should I implement nexus support and open a new PR against develop?

glennmatthews:
Still should go into feature/grpc for now, but we do want it to work on nexus.

I think you can keep nexus compatibility by just checking whether grpc_port is set and branching appropriately.

glennmatthews:
Something like:

```
     when /cisco/
       agentvrf = options[:HOSTS][host.to_s.to_sym]['vrf']
+      grpc_port = options[:HOSTS][host.to_s.to_sym]['grpc_port']
+      if grpc_port.nil?
          return "sudo ip netns exec #{agentvrf} " + cmdstr
+      else
+        node_var = "export NODE=\"127.0.0.1:#{grpc_port} root lab\""
+        return "#{node_var} && ip netns exec #{agentvrf} " + cmdstr
+      end
```

robert-w-gries:
Is there a way to check that the platform is `ios_xr` instead?  Checking `grpc_port.nil?` would fail if using grpc UDS.

glennmatthews:
When we have a UDS, you won't need to set grpc_port. 

This will need to be reworked anyway if/when we have a final design in place from gRPC team.

robert-w-gries:
I meant that this command would fail `return "sudo ip netns exec #{agentvrf} " + cmdstr` since sudo doesn't work right now.  But that bug would probably be fixed before UDS is implemented anyway

I'm going with your suggested change.

robert-w-gries:
Beaker now runs for both nexus and XR

robert-w-gries:
Updated documentation and added comment

PR #78, opened 2015-11-20
Routetarget

andy91315:
1. pass the beaker test
2. pass the static analysis


README.md
```diff
@@ -284,6 +284,9 @@ BGP autonomous system number.  Valid values are String, Integer in ASPLAIN or AS
 ##### `vrf`
 Name of the resource instance. Valid values are string. The name 'default' is a valid VRF representing the global bgp.
 
+##### `route_distinguisher`
+Route Distinguisher (RD) is combined with the IPv4 or IPv6 prefix learned by the PE router to create a globally unique address. Valid values are String, keyword 'default'.
```
```
VPN Route Distinguisher (RD). The RD is combined with the IPv4 or IPv6 prefix learned by the PE router to create a globally unique address. Valid values are a String in the one of the route-distinguisher formats (ASN2:NN, ASN4:NN, or IPV4:NN); the keyword 'auto', or the keyword 'default'.
```

README.md
```diff
@@ -482,6 +485,31 @@ redistribute => [['direct'],
                  ['ospf 3',  'rm_ospf'],
                  ['rip 4']]
 ```
+
+##### `route target both auto`
+Enable/Disable route target both auto only for IBGP. Valid values are true, false, or 'default'.
```
chrisvanheuveln: You need to specify the actual property name here:
```
##### `route_target_both_auto`
(iBGP only) Enable/Disable the route-target 'auto' setting for both import and export target communities. Valid values are true, false, or 'default'.
```

README.md
```diff
@@ -482,6 +485,31 @@ redistribute => [['direct'],
...
+
+##### `route target both auto evpn`
+Enable/Disable route target both auto evpn only for IBGP. Valid values are true, false, or 'default'.
```
chrisvanheuveln: ##### `route_target_both_auto_evpn`
(iBGP only, EVPN only) Enable/Disable the EVPN route-target 'auto' setting for both import and export target communities. Valid values are true, false, or 'default'.
```

File: README.md [hunk 4]
```diff
@@ -482,6 +485,31 @@ redistribute => [['direct'],
...
+
+##### `route target import`
+Set route target import extended communities. Valid values are Array, String, keyword 'default'.
```
```
##### `route_target_import`
Sets the route-target import extended communities. Valid values are an Array or space-separated String of extended communities, or the keyword 'default'.

Examples:

~~~ruby
route_target_import => ['1.2.3.4:5', '33:55']
route_target_export => '4:4 66:66'
\~~~
```

lib/puppet/type/cisco_bgp.rb
```diff
@@ -169,6 +170,23 @@ def name
   # Properties #
   ##############
 
+  newproperty(:route_distinguisher) do
+    desc "Route Distinguisher (RD) is combined with the IPv4 or IPv6 prefix
+          learned by the PE router to create a globally unique address.
+          Valid values are String, keyword 'default'."
```
chrisvanheuveln: Use the updated description from the README instead.

lib/puppet/type/cisco_bgp_af.rb
```diff
@@ -459,6 +469,112 @@ def insync?(is)
...
+  newproperty(:route_target_import, array_matching: :all) do
+    desc "Set route target import extended communities.
+          Valid values are Array, String, keyword 'default'"
+
+    match_error = 'must be specified in AS:nn or IPv4:nn notation'
+
+    validate do |community|
+      community.split.each do |value|
+        fail "Confederation peer value '#{value}' #{match_error}" unless
+          /^(?:\d+\.\d+\.\d+\.)?\d+.\d+$/.match(value) ||
```
chrisvanheuveln: This regex isn't quite right. The `.` is a match any single character so it will match on `111`, `1X1`, etc. Use this instead:

`/^(?:\d+\.\d+\.\d+\.)?\d+:\d+$/`

chrisvanheuveln:
Please update CHANGELOG.md as well. These are just a new properties for existing providers so they need to go under the `Added` section like so:

```
### Added
- Extended cisco_bgp with the following attributes:
  - route_distinguisher
- Extended cisco_bgp_af with the following attributes:
  - route_target_both_auto, route_target_both_auto_evpn
  - route_target_import, route_target_import_evpn
  - route_target_export, route_target_export_evpn
```

PR #79, opened 2015-11-20
Beaker passes on XR for builtin service provider

robert-w-gries:
The following testcases pass for XR and Nexus:

```
fileservicepkg/file_provider_nondefaults.rb
fileservicepkg/service_provider_nondefaults.rb
fileservicepkg/package_provider_nondefaults_1.rb
```

Tested service is now `bootlogd` because `syslog` does not function on XR.


RichWellum:
:+1:

PR #81, opened 2015-11-23
1. Beaker tests for vxlan_global, 2. Fix for idempotence fail in the …

alok-aggarwal:
…provider file

1. Beaker tests passed, 
2. No rubocop errors

### DIFF
--- a/lib/puppet/provider/cisco_vxlan_global/nxapi.rb
+++ b/lib/puppet/provider/cisco_vxlan_global/nxapi.rb
@@ -46,7 +46,7 @@ Puppet::Type.type(:cisco_vxlan_global).provide(:nxapi) do
 
   def initialize(value={})
     super(value)
-    @vxlan_global = Cisco::VxlanGlobal.new
+    @vxlan_global = Cisco::VxlanGlobal.new if Cisco::VxlanGlobal.enabled
     @property_flush = {}
   end
 
--- a/tests/beaker_tests/vxlan_global/test_vxlan_global.rb
+++ b/tests/beaker_tests/vxlan_global/test_vxlan_global.rb
@@ -154,6 +154,9 @@ end
 def build_manifest_vxlan_global(tests, id)
   if tests[id][:ensure] == :absent
     state = 'ensure => absent,'
+    # No need to check for 'ensure' => 'absent' here as vxlan_global doesn't
+    # have instances and puppet resource command is issued with no specific
+    # resource name.
     tests[id][:resource] = {}
   else
     state = 'ensure => present,'
@@ -199,6 +202,7 @@ test_name "TestCase :: #{testheader}" do
   # -------------
   id = 'preclean'
   tests[id][:desc] = 'Preclean'
+  tests[id][:ensure] = :absent
   test_harness_vxlan_global(tests, id)
 
   # -------------------------------------------------------------------
--- a/tests/beaker_tests/vxlan_global/test_vxlan_global.rb
+++ b/tests/beaker_tests/vxlan_global/test_vxlan_global.rb
@@ -1,0 +1,222 @@
+###############################################################################
+# Copyright (c) 2015 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+###############################################################################
+# TestCase Name:
+# -------------
+# test-vxlan_global.rb
+#
+# TestCase Prerequisites:
+# -----------------------
+# This is a Puppet vxlan_global resource testcase for Puppet Agent on
+# Nexus devices.
+# The test case assumes the following prerequisites are already satisfied:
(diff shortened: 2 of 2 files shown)


lib/puppet/provider/cisco_vxlan_global/nxapi.rb
```diff
@@ -46,7 +46,7 @@
 
   def initialize(value={})
     super(value)
-    @vxlan_global = Cisco::VxlanGlobal.new
+    @vxlan_global = Cisco::VxlanGlobal.new if Cisco::VxlanGlobal.enabled
```
mikewiebe: What is the reason for the change?

tests/beaker_tests/vxlan_global/test_vxlan_global.rb
```diff
@@ -0,0 +1,222 @@
...
+  manifest_props: '',
+  code:           [0, 2],
+  resource_props: {},
+}
+
+tests['default_properties'] = {
+  manifest_props: "
+    dup_host_ip_addr_detection_host_moves     => 'default',
+    dup_host_ip_addr_detection_timeout        => 'default',
+    anycast_gateway_mac                       => 'default',
```
mikewiebe: Does this mean that anycast_gateway_mac is not configured by default and is therefore not displayed in the `puppet resource` output?

lib/puppet/provider/cisco_vxlan_global/nxapi.rb
alok-aggarwal: ensure: absent Idempotence beaker test for default properties was always having to remove the resource. Thats why the test was failing with exit code 2, whereas the idempotence test expects always 0.

The reason was in provider initialize, a new instance was always getting created even the resource was disabled and therefore the ensure:absent always removed it which was incorrect. This check fixes it.

tests/beaker_tests/vxlan_global/test_vxlan_global.rb
```diff
@@ -0,0 +1,222 @@
...
+# Full command string for puppet resource command
+def puppet_resource_cmd
+  cmd = UtilityLib::PUPPET_BINPATH + 'resource cisco_vxlan_global'
+  UtilityLib.get_namespace_cmd(agent, cmd, options)
+end
+
+def build_manifest_vxlan_global(tests, id)
+  if tests[id][:ensure] == :absent
+    state = 'ensure => absent,'
+    tests[id][:resource] = {}
```
mikewiebe: Should this be `tests[id][:resource] = { 'ensure' => 'absent' }` ?

tests/beaker_tests/vxlan_global/test_vxlan_global.rb
alok-aggarwal: Thats correct. anycast_gateway_mac doesn't have a defualt value. The default means to unconfigure it.

lib/puppet/provider/cisco_vxlan_global/nxapi.rb
chrisvanheuveln: @mikewiebe We worked through this during consult hours. The problem is that the NU code will instantiate (by default) when you call VxlanGlobal.new; so self.instances gets a new instance every time, meaning ensure => :present is true; meaning destroy gets called every time, breaking idempotence.

tests/beaker_tests/vxlan_global/test_vxlan_global.rb
alok-aggarwal: I think state = 'ensure => absent,' is doing the same.

chrisvanheuveln: @alok-aggarwal It's subtle but they're different. 
state is specifying the manifest entry, while resource specifies the puppet resource output (which needs 'absent' in quotes)

alok-aggarwal: well I used the template file where it is { }.

Adding tests[id][:resource] = { 'ensure' => 'absent' } gives me a failure.
Failed Tests Cases:
      Test Case vxlan_global/test_vxlan_global.rb reported: #<Beaker::DSL::Outcomes::FailTest: 
    TestStep :: Match (?-mix:ensure +=> +'?absent'?) :: FAIL
    >

Do i need to make a change elsewhere ?

mikewiebe: What output do you get from the `puppet resource cisco_vxlan_global` on your nx device when the resource is absent?

alok-aggarwal: Just blank.

root@agent-lab1-nx#puppet resource cisco_vxlan_global

root@agent-lab1-nx#

mikewiebe: @alok-aggarwal Ok, then that is the problem.  I think we can leave the code as `tests[id][:resource] = {}` but please gather the output of the beaker test with `--log-level trace` and make sure the test is issuing the following command `puppet resource cisco_vxlan_global` with no optional args.

alok-aggarwal: yes it does.

mikewiebe:
@alok-aggarwal Changes look good, but I just realized that you have not yet updated the changelog and readme file.  

* https://github.com/cisco/cisco-network-puppet-module/blob/feature/vxlan/CHANGELOG.md#new-feature-support
  * Add an entry under `New feature support`
* https://github.com/cisco/cisco-network-puppet-module/blob/feature/vxlan/README.md#resource-reference
  * Add an entry for your provider along with attributes under both `Resource Type Catalog (by Technology)` and `Cisco Resource Type Catalog (by Name)`

alok-aggarwal:
Done.

PR #82, opened 2015-11-23
add vni to cisco_vrf

andy91315:
1. pass beaker test
 Total Suite Time: 98.02 seconds
      Average Test Time: 98.02 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1
I found the non_default test the non_default value, default value. idempotence and title patten test.
So I add the vni as a property in corresponding place.

2. pass static analysis

### DIFF
--- a/lib/puppet/type/cisco_vrf.rb
+++ b/lib/puppet/type/cisco_vrf.rb
@@ -82,16 +82,9 @@ Puppet::Type.newtype(:cisco_vrf) do
 
   newproperty(:vni) do
     desc "Specify virtual network identifier. Valid values are
-          integers between 4096 and 16777215, or keyword
-          'default' to disable this property"
+          integers or keyword 'default' to disable this property"
     munge do |value|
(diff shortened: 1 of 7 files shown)


lib/puppet/type/cisco_vrf.rb
```diff
@@ -78,4 +79,19 @@ def self.title_patterns
...
+
+  newproperty(:vni) do
+    desc "Specify virtual network identifier. Valid values are
+          integers between 4096 and 16777215, or keyword
+          'default' to disable this property"
+    munge do |value|
+      value = :default if value == 'default'
+      unless value == :default
+        value = value.to_i
+        fail 'vni value should be between 4096 and 16777215' unless
```
mikewiebe: In general I don't think we want to range check like this.  We should simply allow the device to reject invalid ranges.

PR #85, opened 2015-11-24
Update README.md

alok-aggarwal:
Added attributes for vxlan_global.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ The following resources include cisco types and providers along with cisco provi
 * VRF Type
   * [`cisco_vrf`](#type-cisco_vrf)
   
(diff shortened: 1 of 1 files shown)


mikewiebe:
@alok-aggarwal Can you close this PR and resubmit against develop?  Now that we have collapsed feature/vxlan into develop we should not use feature/vxlan any more.  I plan to delete it today.

alok-aggarwal:
@mikewiebe : Is there a quick way of porting this commit to develop or will I have to retype it ?

mikewiebe:
@alok-aggarwal Just cancel this and then go back under your fork and select develop instead of feature/vlxan

PR #87, opened 2015-11-25
Rubocop 0.35.1 passes

robert-w-gries:
```
~/cisco-network-puppet-module$ rubocop
warning: parser/current is loading parser/ruby22, which recognizes
warning: 2.2.3-compliant syntax, but you are running 2.2.1.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
Inspecting 215 files
.......................................................................................................................................................................................................................
```

### DIFF
--- a/docs/template-provider-router.rb
+++ b/docs/template-provider-router.rb
@@ -41,10 +41,10 @@ Puppet::Type.type(:cisco_X__RESOURCE_NAME__X).provide(:nxapi) do
   # Property symbol arrays for method auto-generation. There are separate arrays
   # because the boolean-based methods are processed slightly different.
   X__CONSTANT_NAME__X_NON_BOOL_PROPS = [
-    :X__PROPERTY_INT__X,
+    :X__PROPERTY_INT__X
   ]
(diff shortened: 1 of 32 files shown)


docs/template-provider-router.rb
```diff
@@ -41,10 +41,10 @@
   # Property symbol arrays for method auto-generation. There are separate arrays
   # because the boolean-based methods are processed slightly different.
   X__CONSTANT_NAME__X_NON_BOOL_PROPS = [
-    :X__PROPERTY_INT__X,
+    :X__PROPERTY_INT__X
```
chrisvanheuveln: Ugh. I thought we had a cop that enforced commas on the last entry? Are we going away from that?

robert-w-gries:
```
# Make it neater to extend multi-line arrays and hashes
Style/TrailingComma:
  EnforcedStyleForMultiline: comma
```

I believe this only applies to arrays with multiple entries

chrisvanheuveln: yup, makes sense. 

:+1:

PR #94, opened 2015-12-01
Added NXOS provider for snmp_user

jonnytpuppet:


jonnytpuppet:
tests/beaker_tests/snmp_user/snmp_user_provider_defaults.rb passed in 38.45 seconds
      Test Suite: tests @ 2015-12-01 16:17:57 +0000

jonnytpuppet:
tests/beaker_tests/snmp_user/snmp_user_provider_defaults.rb passed in 51.79 seconds
      Test Suite: tests @ 2015-12-01 16:41:08 +0000

lib/puppet/provider/snmp_user/nxapi.rb
```diff
@@ -0,0 +1,142 @@
...
+
+Puppet::Type.type(:snmp_user).provide(:nxapi) do
+  desc 'The Cisco NXAPI provider for snmp_user.'
+
+  confine feature: :cisco_node_utils
+  defaultfor operatingsystem: :nexus
+
+  mk_resource_methods
+
+  REQUIRES_AUTH_PROPS = [:password, :privacy, :private_key, :localized_key, :engine_id, :auth]
```
hunsbea: auth requires auth?

lib/puppet/provider/snmp_user/nxapi.rb
```diff
@@ -0,0 +1,142 @@
...
+    @property_flush[:ensure] = :absent
+  end
+
+  def validate
+    unless @resource[:auth]
+      invalid = []
+      REQUIRES_AUTH_PROPS.each do |prop|
+        invalid << prop if @resource[prop]
+      end
+      fail ArgumentError, "You must specify the 'auth' property when specifying any of the following properties: #{invalid}" unless invalid.empty?
```
hunsbea: nit: not sure why rubocop doesn't complain about this line length

lib/puppet/provider/snmp_user/nxapi.rb
```diff
@@ -0,0 +1,142 @@
...
+# limitations under the License.
+
+require 'cisco_node_utils' if Puppet.features.cisco_node_utils?
+begin
+  require 'puppet_x/cisco/autogen'
+rescue LoadError # seen on master, not on agent
+  # See longstanding Puppet issues #4248, #7316, #14073, #14149, etc. Ugh.
+  require File.expand_path(File.join(File.dirname(__FILE__), '..', '..', '..',
+                                     'puppet_x', 'cisco', 'autogen.rb'))
+end
```
hunsbea: remove unused autogen requires

lib/puppet/provider/snmp_user/nxapi.rb
```diff
@@ -0,0 +1,142 @@
...
+      privacy:     v.priv_protocol,
+      private_key: v.priv_password,
+    }
+    new(current_state)
+  end # self.properties_get
+
+  def self.instances
+    snmpusers = []
+    Cisco::SnmpUser.users.each do |snmpuser_name, v|
+      snmpusers << properties_get(snmpuser_name.split(' ')[0], v)
```
hunsbea: I found a problem and I think it's related to name not being enough to uniquely identify the resource.

The following user (with engineID configured) exists:
snmp-server user foo auth sha 0x7e5030ffd26d7e1b366a9041e9c63c94 priv aes-128 0xcc012f26b3384d4b3da979bff48b4ffd localizedkey engineID 22:22:22:22:22:22

And I want to configure this user (no engineID):
  snmp_user { "foo":
    ensure          => present,
    roles           => ['network-admin'],
    auth            => 'sha',
    password        => '0x7e5030ffd26d7e1b366a9041e9c63c94',
    privacy         => 'aes128',
    private_key     => '0xcc012f26b3384d4b3da979bff48b4ffd',
    localized_key   => true,
}

What happens is that this CLI string gets sent:
snmp-server user foo network-admin auth sha 0x7e5030ffd26d7e1b366a9041e9c63c94 priv aes-128 0xcc012f26b3384d4b3da979bff48b4ffd localizedkey engineID 22:22:22:22:22:22

And it fails to add the network-admin group because this and engineID are mutually exclusive. I noticed it doesn't call destroy on the resource before trying to update. My theory is that you need engineID as part of the name to uniquely identify the resource and properly remove the existing user first.

tests/beaker_tests/snmp_user/snmp_userlib.rb
```diff
@@ -0,0 +1,99 @@
...
+    manifest_str = "cat <<EOF >#{UtilityLib::PUPPETMASTER_MANIFESTPATH}
+node default {
+    snmp_user { 'test_snmp_user':
+     ensure          => absent,
+  }
+}
+EOF"
+    manifest_str
+  end
+end
```
hunsbea: I'd really like to see a test for engine_id.

Would be nice: negative test trying to set both engine_id and roles, maybe a test setting auth without priv, or a negative test for the vice versa

jonnytpuppet: @hunsbea To the best of my knowledge, we haven't done negative testing in any of our other netdev providers that we have implemented (and have been merged). It's definitely a nice to have, but from a deliverable standpoint we've been focusing on CRUD tests, in line with the "acceptance" testing done in other Puppet modules.

lib/puppet/provider/snmp_user/nxapi.rb
jonnytpuppet: @hunsbea Well spotted bug. However, the issue isn't uniqueness per se. The issue lies in the fact that the provider will use the values obtained from the currently configured resource. I need to think of a way to only use the puppet-provided values and ignore the currently set items.

I don't believe it's valid to have 1 user configured twice on the system? This doesn't really make sense?

jonnytpuppet:
@hunsbea Most concerns have been addressed.

Unfortunately, this provider will only support a single user, as that is what it keys on. This is a limitation of the netdev provider and the validation it does on the namevar.

After discussion internally, it is generally accepted that it is not good practise to manage the same resource using Puppet and manually. After a few bug fixes (with a test added to make sure that a change from roles to engine_id works), everything should work ok if using just this provider within it's limitations.

jonnytpuppet:
tests/beaker_tests/snmp_user/snmp_user_provider_defaults.rb passed in 41.39 seconds
      Test Suite: tests @ 2015-12-02 14:59:27 +0000

hunsbea:
@jonnytpuppet I retested, and I see my engineID example is still a problem. The use case is not that the user is managing two different users or managing two different ways. The problem is that if someone had previously configured a user using an engineID (using puppet, for example), the provider is unable to update the user to remove the engineID and configure a user with a role later. In my opinion this is a bug in basic functionality that must be addressed.

jonnytpuppet:
@hunsbea I thought I had already addressed that in the latest push. Can you please confirm you're using the latest code?

jonnytpuppet:
@hunsbea I just tested using the latest code and your CLI command and manifest, and it works for me....

jonnytpuppet:
```
switch(config)# snmp-server user foo auth sha 0x7e5030ffd26d7e1b366a9041e9c63c94 priv aes-128 0xcc012f26b3384d4b3da979bff48b4ffd localizedkey engineID 22:22:22:22:22:22
switch(config)# show running-config snmp all | include foo
snmp-server user foo auth sha 0x7e5030ffd26d7e1b366a9041e9c63c94 priv aes-128 0xcc012f26b3384d4b3da979bff48b4ffd localizedkey engineID 22:22:22:22:22:22

Linux# /opt/puppetlabs/puppet/bin/puppet resource snmp_user foo
snmp_user { 'foo':
  ensure      => 'present',
  auth        => 'sha',
  engine_id   => '22:22:22:22:22:22',
  password    => '0x7e5030ffd26d7e1b366a9041e9c63c94',
  privacy     => 'aes128',
  private_key => '0xcc012f26b3384d4b3da979bff48b4ffd',
}

Linux# cat h.pp
snmp_user { "foo":
    ensure          => present,
    roles           => ['network-admin'],
    auth            => 'sha',
    password        => '0x7e5030ffd26d7e1b366a9041e9c63c94',
    privacy         => 'aes128',
    private_key     => '0xcc012f26b3384d4b3da979bff48b4ffd',
    localized_key   => true,
}

Linux# /opt/puppetlabs/puppet/bin/puppet apply h.pp
Notice: Compiled catalog for 10.168.1.3 in environment production in 0.35 seconds
Notice: /Stage[main]/Main/Snmp_user[foo]/roles: roles changed '[]' to '['network-admin']'
Notice: Applied catalog in 2.42 seconds
Linux# /opt/puppetlabs/puppet/bin/puppet resource snmp_user foo
snmp_user { 'foo':
  ensure      => 'present',
  auth        => 'sha',
  password    => '0x7e5030ffd26d7e1b366a9041e9c63c94',
  privacy     => 'aes128',
  private_key => '0xcc012f26b3384d4b3da979bff48b4ffd',
  roles       => ['network-admin'],
}
Linux#

**Comment by hunsbea** [comment]
@jonnytpuppet did you push your latest changes? I only see one commit in the pull request history

**Comment by jonnytpuppet** [comment]
@hunsbea Yep, I sure have. The commit ID is 8b6b087b0a17fdec005ec8359c20de13abc3cbf4

```
$ git push --force jonnytpuppet snmp_user
Everything up-to-date

hunsbea:
@jonnytpuppet yep my bad, misconfig. thanks for making those changes, merging now

PR #96, opened 2015-12-03
channel_group addition to cisco_interface

saichint:


README.md
```diff
@@ -760,6 +760,13 @@ Shutdown state of the interface. Valid values are 'true', 'false', and
 Switchport mode of the interface. To make an interface Layer 3, set
 `switchport_mode` to 'disabled'. Valid values are 'disabled', 'access', 'tunnel', 'fex_fabric', 'trunk', and 'default'.
 
+###### `channel_group`
+channel_group is an aggregation of multiple physical interfaces
+that creates a logical interface. Valid values are 1 to 4096 and 'default'
+When we delete channel_group from an interface, the port-channel interface
+will not be removed. If this needs to be removed, use the
+interface destroy method on the port-channel
```
chrisvanheuveln: 1. This is a Markdown file so the text shouldn't be wrapped. It's easier to update if it's not wrapped.

2. The interface properties are listed as common, L2, or L3 properties. If channel-group is L2 only, move it there, or whereever it needs to go.

3. Please keep the property names in alphabetic order. (I see a couple of others are out of order but I'll fix that)

4. Try to keep the 'we' references out of the documents. Your description is good but I would probably reword it slightly to this:

```
###### `channel_group`
channel_group is an aggregation of multiple physical interfaces that creates a logical interface. Valid values are 1 to 4096 and 'default'. 

Note: On some platforms, a normal side-effect of adding the channel-group property is that an independent port-channel interface will be created; however, removing the channel-group configuration by itself will not also remove the port-channel interface. Therefore, the port-channel interface itself may be explicitly removed by using the `cisco_interface` provider with `ensure => absent`.
```

chrisvanheuveln:
Sai, I ran the beaker test and it looks like we have an issue on some platforms -- if you try to remove channel-group when it's not present then the cli raises an error:

```
 /isan/bin/vsh -c 'conf t ; interface eth1/4 ; no channel-group'
Last 10 lines of output were:
        Ethernet1/4: not part of port-channel>
```

What surprises me though is that puppet is not detecting that it's not present (if it did then it would not even try to call the setter). I suspect that we need to change the yaml default entry to ~ (nil) like this:

```
  default_value: ~
```

You can do some manual `puppet agent -t` tests after changing that and see if it works. Just set your manifest up with `channel_group => 'default'`, make sure your interface does not have channel-group on it already, then do `puppet agent -t`.  If that doesn't fix it I'll look at it again next week.

chrisvanheuveln:
I fixed the NU side last week and the rubocop issue this morning. There is a small commit I need to do for beaker that I'll take care of right after this merge.

chrisvanheuveln:
I'm having trouble getting the merge to work so I'll close this and re-open it under my fork.

PR #98, opened 2015-12-04
Updated beaker tests for cisco_interface

robert-w-gries:
Beaker tests for cisco_interface now use the common test harness.

All tests pass on both XR and Nexus.

## New property hashes

Usually, we can avoid explicitly setting both `manifest_props` and `resource_props` because those hashes contain redundant information.  In most cases, we can just specify a `key` (the property we want to set) and a `value` (the expected value of the property).  This condenses the property information from `manifest_props` and `resource_props` into one hash.

I added two new hashes to the beaker test: `default_values` and `non_default_values`.  When the test builds the `cisco_interface` manifest, it adds all key-value pairs from `default_values` and `non_default_values` to both the `manifest_props` string and `resource` hash.

### Default Values Hash Behavior

* Append `\n#{key} => 'default'` to `manifest_props` string
* Remove key from `default_values` hash if default value causes absence in puppet resource command
* Merge remaining key-value pairs in `default_values` hash into `resource` hash

### Non-default Values Hash Behavior

This hash is used whenever the input value matches the output value.

* Append `\n#{key} => #{value}` to `manifest_props` string
* Merge key-value pairs in `non_default_values` hash into `resource` hash


tests/beaker_tests/cisco_interface/test_l3_ethernet.rb
```diff
@@ -0,0 +1,356 @@
...
+    },
+  }
+
+  if platform == 'nexus'
+    tests['default_properties'][:non_default_values].merge!(
+      'switchport_mode' => "'disabled'",
+    )
+    tests['default_properties'][:default_values].merge!(
+      'speed'                        => "'auto'",
+      'duplex'                       => "'auto'",
```
glennmatthews: This feels a bit off to me. No doubt this is because of the fact that we're interacting with the Puppet CLI, so it's easier to treat everything as a string, and strings as double-strings, but it would be nice to be able to leave off the quoting and define things as their actual types (and have the helper function handle the stringification as needed):
```ruby
    tests['default_properties'][:default_values].merge!(
      'speed'                        => 'auto',
      'duplex'                       => 'auto',
      'switchport_autostate_exclude' => false,
      'switchport_vtp'               => false,
    )
```

Something for a future US, perhaps.

tests/beaker_tests/cisco_interface/test_l3_ethernet.rb
```diff
@@ -0,0 +1,356 @@
...
+  node 'default' {
+    cisco_interface { '#{testcase[:title_pattern]}':
+      #{state}
+      #{testcase[:manifest_props]}
+    }
+  }
+EOF"
+end
+
+def invalid_intf?(interface)
```
glennmatthews: perhaps `nonensurable_intf?` would be more clear?

robert-w-gries: How about `invalid_absent_intf?`?

glennmatthews: Fine by me.

tests/beaker_tests/cisco_interface/test_l3_ethernet.rb
robert-w-gries: I took another look at removing the extra quotations, and it ended up being a fairly easy fix.

robert-w-gries:
I re-tested after my latest changes, and I've been running into idempotency issues with switchport properties.  It also looks like there are some more switchport properties that need to be properly tested.

I also ran into an issue with the negative test where it shut down my XR's management port.  Looking into that as well

robert-w-gries:
All tests now pass for both platforms.

* [Small commit](https://github.com/cisco/cisco-network-node-utils/commit/e041e8b185330ca8dfb0dbce5f0261db62386887) to node-utils fixes switchport_trunk_allowed_vlan default value.

* Switchport_vtp property issue was fixed by configuring switch with `feature vtp`.  Unable to reproduce beaker issue afterwards.

* Negative test is working after rebuilding puppet module.  Not sure what was causing the issue.

One more issue that might need to be addressed by this PR is cleaning up before and after tests.  Any thoughts on how it should be handled?

chrisvanheuveln:
@robert-w-gries *cleaning up before and after tests* 

Mike and I had discussed using `puppet resource \<provider>` to get current state and then follow it with `puppet resource \<provider> ensure=absent` to cleanup. Will that work for you?

glennmatthews:
Shouldn't even need to do the former, I would think - for virtual interfaces, we can just `ensure=absent` and handle both the 'success, change' and 'success, no change' return codes. For physical interfaces (Ethernet) that we can't `ensure=absent`, I think we'd talked about just explicitly setting all supported attributes to `default` as the preclean/postclean.

chrisvanheuveln:
@glennmatthews @robert-w-gries 

Agreed, no need to check the former for interfaces, I was speaking generally. Alex already has a helper method coded to what we're talking about:
https://github.com/cisco/cisco-network-puppet-module/pull/100/files#diff-e0883d3270eb11e5d50fcdc5c4f83003R309

robert-w-gries:
@chrisvanheuveln I went ahead with manually specifying the cleanup for this test because I'm unable to use `ensure=absent` for nexus physical interfaces.

robert-w-gries:
Updated docs.

docs/template-beaker-router.rb
```diff
@@ -65,79 +64,83 @@
...
-  ",
-  resource_props: {
-    'shutdown' => 'true'
-  },
-}
+  tests['default_properties'] = {
+    title_pattern:  '1',
+    # Input:  maximum_paths => 'default',
+    #         shutdown      => 'default',
+    # Output: maxium_paths  => 8,
```
glennmatthews: s/maxium/maximum/g

glennmatthews:
:+1:

PR #100, opened 2015-12-06
aaa authorization login config service

hunsbea:
validated against rubocop 35.1, beaker passes, etc


tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -291,6 +291,39 @@ def test_idempotence(tests, id)
...
+# @param res_name [String] the resource name that will be cleaned up
+def manifest_absent_cleanup(agent, res_name, stepinfo='absent clean')
+  # Don't forget to UtilityLib.set_manifest_path(master, self)
+  step "TestStep :: #{stepinfo}" do
+    # create an absent resource for each instance within node default
+    abs_str = ''
+    get_current_resource_instances(agent, res_name).each do |title|
+      abs_str << "#{res_name} { '#{title}':\n    ensure => absent,\n  }\n"
+    end
+    manifest_str = "cat <<EOF >#{UtilityLib::PUPPETMASTER_MANIFESTPATH}\n" \
```
mikewiebe: Instead of applying a manifest I think @chrisvanheuveln 's suggestion is better.

Mike and I had discussed using puppet resource <provider> to get current state and then follow it with puppet resource <provider> ensure=absent to cleanup.

lib/puppet/provider/cisco_aaa_authorization_login_cfg_svc/nxapi.rb
```diff
@@ -0,0 +1,136 @@
...
+  defaultfor operatingsystem: :nexus
+
+  mk_resource_methods
+
+  AAA_CFG_PROPS = [:method]
+  PuppetX::Cisco::AutoGen.mk_puppet_methods(:non_bool, self, '@aaa_login_svc',
+                                            AAA_CFG_PROPS)
+
+  def initialize(value={})
+    super(value)
```
mikewiebe: Add comments here to make it clear that this class method is shared.

lib/puppet/type/cisco_aaa_authorization_login_cfg_svc.rb
```diff
@@ -0,0 +1,113 @@
...
+  cisco_aaa_authorization_login_cfg_svc {\"[console|default]\":
+    ..attributes..
+  }
+~~~
+
+  'console' and 'default' are the only services configurable.
+
+  Example:
+~~~puppet
+    cisco_aaa_authorization_login_cfg_svc {\"console\":
```
mikewiebe: *ni* Don't think you need the `\` since this is all documentation.

tests/beaker_tests/aaalogincfgsvc/test_aaalogincfgsvc_defaults.rb
```diff
@@ -0,0 +1,137 @@
...
+#
+# The testcode also uses RegExp pattern matching on stdout or output IO
+# instance attributes to verify resource properties.
+#
+###############################################################################
+# rubocop:disable Style/HashSyntax,Style/ExtraSpacing
+
+# Require UtilityLib.rb and AaaLoginCfgSvcLib.rb paths.
+require File.expand_path('../../lib/utilitylib.rb', __FILE__)
+require File.expand_path('../aaalogincfgsvclib.rb', __FILE__)
```
mikewiebe: What is the rational for a separate lib file? Most of our new beaker test's are just combined into a single file with the new format.

lib/puppet/type/cisco_aaa_authorization_login_cfg_svc.rb
```diff
@@ -0,0 +1,113 @@
...
+    validate do |value|
+      fail unless value == 'default' || value == 'console'
+    end
+  end
+
+  ##############
+  # Attributes #
+  ##############
+
+  newproperty(:groups, array_matching: :all) do
```
mikewiebe: For properties that use a single dimensional arrays we have started support both of the following with the hope that it's somewhat more user friendly when creating the manifest.

1) ['string1', 'string2']
2) "string1 string2"

Here is an example of the type code needed to support both.

https://github.com/cisco/cisco-network-puppet-module/blob/develop/lib/puppet/type/cisco_bgp_af.rb#L573

Only minor changes needed to split and process.

lib/puppet/type/cisco_aaa_authorization_login_cfg_svc.rb
```diff
@@ -0,0 +1,113 @@
...
+
+    newvalues(:local, :unselected, :default)
+  end
+
+  ################
+  # Autorequires #
+  ################
+
+  # Autorequire all cisco_aaa_group_tacacs associated with this service
+  autorequire(:cisco_aaa_group_tacacs) do |rel_catalog|
```
mikewiebe: Do we need beaker tests for this?

mikewiebe:
@hunsbea General comment to refactor node_utils to remove internal tacacs server reference

https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/test_aaa_authorization_service.rb#L38

tests/beaker_tests/aaalogincfgsvc/test_aaalogincfgsvc_defaults.rb
hunsbea: The defaults, nondefaults, and negative tests all share the manifest creation functions, so I think it still makes sense to put them in a shared file.

lib/puppet/type/cisco_aaa_authorization_login_cfg_svc.rb
hunsbea: Just for future reference, this change also required changes in the provider (adding a flatten in the setter), and also adding a flatten in the autorequire in the type.

lib/puppet/type/cisco_aaa_authorization_login_cfg_svc.rb
hunsbea: I reversed the order of the manifest so if my autorequires don't work, the tests will fail.

mikewiebe:
@hunsbea Update CHANGELOG and README

hunsbea:
updated docs :)

CHANGELOG.md
```diff
@@ -39,6 +39,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### New feature support
 #### Cisco Resources.
+- cisco_aaa_authentication_login type and provider.
```
mikewiebe: This is in the wrong section :)  It's under release 1.1.0 section

hunsbea: oops, moved

PR #101, opened 2015-12-08
BGP Neighbor puppet support for XR.

chburket:
New PR to fix the commit history of the last one.  No changes since last review.

### DIFF
--- a/examples/demo_bgp.pp
+++ b/examples/demo_bgp.pp
@@ -130,46 +130,72 @@ class ciscopuppet::demo_bgp {
   #---------------------------------------------------------------------------#
   # Configure BGP IPv4 Neighbors
   #---------------------------------------------------------------------------#
+
+  $capability_negotiation = $operatingsystem ? {
+    'nexus' => true,
+    default => undef
+  }
+  $dynamic_capability = $operatingsystem ? {
+    'nexus' => true,
+    default => undef
+  }
+  $log_neighbor_changes = $operatingsystem ? {
+    'nexus' => disable,
+    default => undef
+  }
+  $low_memory_exempt = $operatingsystem ? {
+    'nexus' => false,
+    default => undef
+  }
+  $remove_private_as = $operatingsystem ? {
+    'nexus' => 'all',
+    default => undef
+  }
+  $update_source = $operatingsystem ? {
+    'nexus' => 'ethernet1/1',
+    default => 'fastethernet1/1/1/1'
+  }
+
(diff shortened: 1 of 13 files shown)


examples/demo_bgp.pp
```diff
@@ -130,46 +130,72 @@
...
     remote_as              => 12,
-    remove_private_as      => 'all',
+    remove_private_as      => $remove_private_as,
     shutdown               => true,
     suppress_4_byte_as     => true,
     timers_keepalive       => 90,
     timers_holdtime        => 270,
-    update_source          => 'ethernet1/1',
+    update_source          => $update_source,
     transport_passive_only => false,
```
glennmatthews: Probably should add an example of transport_passive_mode to the demo manifest?

tests/beaker_tests/bgpneighbor/test_bgpneighbor_defaults.rb
```diff
@@ -87,15 +89,19 @@ def create_bgp_neighbor_defaults(tests, id, title, string=false)
...
     'ebgp_multihop'          => 'false',
     'local_as'               => '0',
-    'low_memory_exempt'      => 'false',
-    'maximum_peers'          => '0',
-    'remote_as'              => '0',
+    'remote_as'              => remote_as.to_s,
     'suppress_4_byte_as'     => 'false',
     'timers_keepalive'       => '60',
     'timers_holdtime'        => '180',
     'transport_passive_only' => 'false',
```
glennmatthews: Beaker tests should probably also exercise transport_passive_mode?

glennmatthews:
Besides the above questions, :+1:

examples/demo_bgp.pp
chburket: Since the main point of this story was functional equivalence to NX, I was going to just leave transport_passive_mode as an unadvertised feature until the story where we add XR-specific properties.  I can be talked into fully documenting it and adding it to the beaker tests, though.  (I did test it through beaker, I just removed it before I committed).

glennmatthews: I'm not a big fan of undocumented features - if it's working in XR, let's go ahead and document/test it. If it's not ready for that yet then we shouldn't be committing the code for that feature. IMHO.

Glenn

chburket: @glennmatthews 
Ok, I added transport_passive_mode doc to the README, an example to demo_bgp.pp, and a new beaker test file to cover the property.

PR #103, opened 2015-12-08
Add .rubocop.yml for beaker_tests directories

chrisvanheuveln:


mikewiebe:
Context for this change?

glennmatthews:
I thought we'd considered the idea of disabling Style/ModuleLength across the board as not being particularly useful.

chrisvanheuveln:
@mikewiebe a beaker test method got too long - Sai added 15 lines to test channel-group and that pushed it over the limit.

@glennmatthews Yes we had but I thought the conclusion was to just disable it for test files. I'm fine with turning it off globally. I'll make the change.

PR #107, opened 2015-12-11
Beaker changes for bgp on XR

chburket:


README.md
```diff
@@ -304,25 +304,25 @@ VPN Route Distinguisher (RD). The RD is combined with the IPv4 or IPv6 prefix le
...
 
 ##### `maxas_limit`
-Specify Maximum number of AS numbers allowed in the AS-path attribute. Valid values are integers between 1 and 512, or keyword 'default' to disable this property.
+Specify Maximum number of AS numbers allowed in the AS-path attribute. Valid values are integers between 1 and 512, or keyword 'default' to disable this property. This property is not supported on IOS XR.
 
 ##### `shutdown`
-Administratively shutdown the BGP protocol. Valid values are 'true', 'false', and 'default'.
+Administratively shutdown the BGP protocol. Valid values are 'true', 'false', and 'default'. This property is not supported on IOS XR.
 
 ##### `supress_fib_pending`
```
glennmatthews: Should this instead be `suppress_fib_pending`?

examples/demo_bgp.pp
```diff
@@ -19,7 +19,29 @@
   # --------------------------------------------------------------------------#
   # Configure Global BGP                                                      #
   # --------------------------------------------------------------------------#
-  cisco_bgp { '55.77 blue':
+
+  $maxas_limit = $operatingsystem ? {
+    'nexus' => '50',
+    default => undef
```
glennmatthews: Question for my own curiosity - if we default to `default` instead of `undef`, does that still work with our code?

glennmatthews:
Looks like this was missed previously - puppet/provider/cisco_bgp/nxapi.rb should be renamed to puppet/provider/cisco_bgp/cisco.rb (it's not NXAPI specific any more) and the 'provides' statement in there should similarly change :nxapi to :cisco. Would you mind doing the appropriate 'git mv' and change as part of this PR?

tests/beaker_tests/.rubocop.yml
```diff
@@ -0,0 +1,4 @@
+# Rubocop configuration
+
```
glennmatthews: You need to add `inherit_from: ../.rubocop.yml` so that the configuration in the parent file is inherited into this directory. Once you make that change, Travis-CI should be passing for this PR.

chburket: Shoot.  Forgot that part.  Will fix.

examples/demo_bgp.pp
chburket: Yes, I believe so.  I've seen that no "UnsupportedError" is thrown when setting unsupported properties to "default" (at least some of the properties).

README.md
chburket: Yes.  I'll fix.

chburket:
*All* of them are still named nxapi.rb.  Should we create another story to handle the mass rename?

chburket:
@glennmatthews @RichWellum 

Pushed next round of changes.  I have so far not renamed all the nxapi.rb files (see above).

glennmatthews:
We're not going to rename all of the files at once - just the ones that have been certified to support XR. (NX-only providers should stay 'nxapi' for now, IMO). So this should just be the cisco_command_config, cisco_interface, and cisco_bgp* providers that need to be renamed at this point.

examples/demo_bgp.pp
glennmatthews: We should see that for all of them, actually - setting an unsupported property to 'default' should be a no-op, IIRC. Might be worth trying that in the demo manifest and/or the beaker tests, come to think of it.

tests/beaker_tests/bgp/bgplib.rb
```diff
@@ -360,8 +283,7 @@ def self.create_bgp_manifest_title_pattern1
         ensure                                 => present,
         asn                                    => #{BgpLib::ASN},
       }
-    }
-    EOF"
+    }\nEOF"
```
glennmatthews: Was this change to make rubocop happy, or...?

glennmatthews:
:+1:

tests/beaker_tests/bgp/bgplib.rb
chburket: The temp manifests were ending up with "   EOF" as the last line, which puppet was ignoring (or warning about, can't remember).  Removing the leading spaces before "EOF" made it work as expected.

glennmatthews: Ah cool, nice catch. Do any other beaker tests need similar fixes?

chburket: vrf/vrflib does.  I fixed bgp neighbor, already.  But I see that I missed 2 spots in bgplib, so I'll update those.

chburket:
@glennmatthews @RichWellum 
Hopefully, last round of changes.

tests/beaker_tests/bgp/bgplib.rb
```diff
@@ -34,22 +34,36 @@ module BgpLib
...
+        "route_distinguisher                    => 'default',"
+    end
+
     manifest_str = "cat <<EOF >#{UtilityLib::PUPPETMASTER_MANIFESTPATH}
     node 'default' {
       cisco_bgp { 'default':
         ensure                                 => present,
         asn                                    => #{BgpLib::ASN},
-        vrf                                    => 'default',
+        vrf                                    => '#{vrf}',
```
RichWellum: Why changing this from default to non-default vrf?

tests/beaker_tests/bgp/bgplib.rb
```diff
@@ -34,22 +34,36 @@ module BgpLib
...
         router_id                              => 'default',
```
RichWellum: Aside from this one comment, it looks :+1: to me.

tests/beaker_tests/bgp/bgplib.rb
chburket: I'm just allowing the vrf to be specified as a method argument, so the code can be shared.  'default' is still the default (so no behavior change if you don't specify a vrf when calling the method).

RichWellum: Seems a little out of place - because the whole example is defaulted. But as the default=default(sorry) I guess it's ok.

chburket: There were three methods that were identical except for the vrf:  create_bgp_manifest_present, create_bgp_manifest_present_vrf1, and create_bgp_manifest_present_vrf2.  The intent of each was to test default values for global bgp, vrf1 ('blue') and vrf2 ('red'), respectively.  The last two methods now simply call the first method with 'blue' and 'red'.

chburket:
@glennmatthews 

Latest round is pushed.

PR #109, opened 2015-12-14
Added NXOS provider for snmp_notification_receiver

jonnytpuppet:
Requires https://github.com/puppetlabs/netdev_stdlib/pull/17 to be merged first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,9 +63,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - ntp_server provider.
 - radius provider.
 - radius global provider.
+- snmp_notification_receiver provider.
+- snmp_user provider.
 - syslog_server provider.
 - syslog_setting provider.
-- snmp_user provider.
 
 ### Added
 - New documentation for developing beaker testcases: README-develop-beaker-scripts.md
--- a/README.md
+++ b/README.md
@@ -199,6 +199,7 @@ The following resources include cisco types and providers along with cisco provi
   * [`network_snmp (netdev_stdlib)`](#type-network_snmp)
   * [`snmp_community (netdev_stdlib)`](#type-snmp_community)
   * [`snmp_notification (netdev_stdlib)`](#type-snmp_notification)
+  * [`snmp_notification_receiver (netdev_stdlib)`](#type-snmp_notification_receiver)
   * [`snmp_user (netdev_stdlib)`](#type-snmp_user)
 
 * SYSLOG Types
@@ -274,6 +275,7 @@ The following resources include cisco types and providers along with cisco provi
 * [`search_domain`](#type-search_domain)
 * [`snmp_community`](#type-snmp_community)
 * [`snmp_notification`](#type-snmp_notification)
+* [`snmp_notification_receiver`](#type-snmp_notification_receiver)
 * [`snmp_user`](#type-snmp_user)
 * [`syslog_server`](#type-syslog_server)
 * [`syslog_setting`](#type-syslog_setting)
@@ -1673,6 +1675,40 @@ Manages an SNMP notification on a Cisco SNMP server.
 ##### `enable`
 Determine whether the trap should be on or off. Valid
 values are true and false.
+
+### Type: snmp_notification_receiver
+
+Manages an SNMP user on an cisco SNMP server.
+
+#### Parameters
+
+##### `ensure`
+Determines whether the config should be present or not on the device. Valid
+values are 'present', and 'absent'.
+
+##### `name`
+IP address of the SNMP user. Valid value is a string.
+
+##### `port`
+SNMP UDP port number
+
+##### `username`
+Username to use for SNMPv3 privacy and authentication.  This is the community string for SNMPv1 and v2.
+
+##### `version`
+SNMP version [v1|v2|v3]
+
+##### `type`
+The type of receiver [traps|informs].
+
+##### `security`
+SNMPv3 security mode [auto|noauth|priv].
+
+##### `vrf`
+Interface to send SNMP data from, e.g. "management"
+
+##### `source_interface`
+Source interface to send SNMP data from, e.g. "ethernet 2/1".
 
 ### Type: snmp_user
 
--- a/lib/puppet/provider/snmp_notification_receiver/nxapi.rb
+++ b/lib/puppet/provider/snmp_notification_receiver/nxapi.rb
@@ -1,0 +1,168 @@
+# The NXAPI provider for cisco snmp_notification_receiver.
+#
+# November, 2014
+#
+# Copyright (c) 2014-2015 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'cisco_node_utils' if Puppet.features.cisco_node_utils?
+
+Puppet::Type.type(:snmp_notification_receiver).provide(:nxapi) do
+  desc 'The Cisco NXAPI provider for snmp_notification_receiver.'
+
+  confine feature: :cisco_node_utils
+  defaultfor operatingsystem: :nexus
+
+  REQUIRED_PROPS = [:type, :version, :username]
+
+  mk_resource_methods
+
+  def initialize(value={})
+    super(value)
+
+    @snmpnotificationreceiver = Cisco::SnmpNotificationReceiver.receivers[name]
+    @property_flush = {}
+    debug 'Created provider instance of snmp_notification_receiver'
+  end
+
+  def self.properties_get(snmpnotificationreceiver_name, v)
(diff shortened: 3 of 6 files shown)


jonnytpuppet:
tests/beaker_tests/snmp_notification_receiver/snmp_notification_receiver_provider_defaults.rb passed in 47.16 seconds
      Test Suite: tests @ 2015-12-14 16:58:35 +0000

mikewiebe:
@jonnytpuppet Will wait for https://github.com/cisco/cisco-network-node-utils/pull/129 to be completed/merged first.

jonnytpuppet:
@mikewiebe Code updated

```
tests/beaker_tests/snmp_notification_receiver/snmp_notification_receiver_provider_defaults.rb passed in 45.11 seconds
      Test Suite: tests @ 2015-12-18 12:56:46 +0000
```

jonnytpuppet:
@mikewiebe @chrisvanheuveln can this be merged please?

mikewiebe:
@jonnytpuppet I just got back from break and have not looked at this code yet.  Will review it today now that the node_utils piece is complete.

mikewiebe:
@jonnytpuppet What is the status of https://github.com/puppetlabs/netdev_stdlib/pull/17?  Do we have an eta on when this will be merged?

CHANGELOG.md
```diff
@@ -66,6 +66,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - syslog_server provider.
 - syslog_setting provider.
 - snmp_user provider.
+- snmp_notification_receiver provider.
```
mikewiebe: *nit* Alpha order.

README.md
```diff
@@ -1674,6 +1676,40 @@ Manages an SNMP notification on a Cisco SNMP server.
...
+Manages an SNMP user on an cisco SNMP server.
+
+#### Parameters
+
+##### `ensure`
+Determines whether the config should be present or not on the device. Valid
+values are 'present', and 'absent'.
+
+##### `name`
+Name of the SNMP user. Valid value is a string.
```
mikewiebe: The type indicates that this is `Hostname or IP address of the receiver`  Which description should be used here?

lib/puppet/provider/snmp_notification_receiver/nxapi.rb
```diff
@@ -0,0 +1,164 @@
...
+    end
+
+    snmpnotificationreceivers
+  end
+
+  def self.prefetch(resources)
+    snmpnotificationreceivers = instances
+
+    resources.keys.each do |id|
+      provider = snmpnotificationreceivers.find { |snmpnotificationreceiver| snmpnotificationreceiver.name.to_s == id.to_s }
```
mikewiebe: Line length.

tests/beaker_tests/snmp_notification_receiver/snmp_notification_receiverlib.rb
```diff
@@ -0,0 +1,101 @@
...
+  def self.create_snmp_notification_receiver_manifest_present
+    manifest_str = "cat <<EOF >#{UtilityLib::PUPPETMASTER_MANIFESTPATH}
+node default {
+  snmp_notification_receiver { '2.3.4.5':
+    ensure           => 'present',
+    source_interface => 'ethernet1/3',
+    port             => '47',
+    type             => 'traps',
+    username         => 'jj',
+    version          => 'v3',
```
mikewiebe: We should make sure we have test coverage for the following.

1) All types
2) All versions
3) All Security options.
4) Set all properties to default.

tests/beaker_tests/snmp_notification_receiver/snmp_notification_receiver_provider_defaults.rb
```diff
@@ -0,0 +1,188 @@
...
+
+# @test_name [TestCase] Executes defaults testcase for snmp_notification_receiver Resource.
+test_name "TestCase :: #{testheader}" do
+  # @step [Step] Sets up switch for provider test.
+  step 'TestStep :: Setup switch for provider' do
+    # Define PUPPETMASTER_MANIFESTPATH constant using puppet config cmd.
+    UtilityLib.set_manifest_path(master, self)
+
+    # Ensure that resource is removed beforehand
+    on(master, SnmpNotificationReceiverLib.create_snmp_notification_receiver_manifest_absent)
```
mikewiebe: Instead of this cleanup logic, we now have a new method that will remove all instances of a particular resource.  Let's use that instead.

Here is an example of how it can be called.

https://github.com/cisco/cisco-network-puppet-module/blob/develop/tests/beaker_tests/cisco_vxlan_vtep/test_vxlan_vtep.rb#L215

lib/puppet/provider/snmp_notification_receiver/nxapi.rb
```diff
@@ -0,0 +1,164 @@
...
+    return if @property_flush[:ensure] == :absent
+
+    validate
+
+    # Needed due to NXOS issue regarding removing and adding resources too quickly
+    sleep 2
+
+    @snmpnotificationreceiver = \
+      Cisco::SnmpNotificationReceiver.new(@resource[:name],
+                                          instantiate:      true,
```
mikewiebe: Line length?

lib/puppet/provider/snmp_notification_receiver/nxapi.rb
```diff
@@ -0,0 +1,164 @@
...
+      '1'
+    elsif @resource[:version].eql?(:v2)
+      '2c'
+    elsif @resource[:version].eql?(:v3)
+      '3'
+    end
+  end
+
+  def flush # rubocop:disable Metrics/CyclomaticComplexity
+    @snmpnotificationreceiver.destroy if @snmpnotificationreceiver
```
mikewiebe: Not sure I understand this.  I would have expected this to be..
 ```
if @property_flush[:ensure] == :absent
  @snmpnotificationreceiver.destroy
  @snmpnotificationreceiver = nil
...
```

Comment (jonnytpuppet) [inline lib/puppet/provider/snmp_notification_receiver/nxapi.rb]: This provider needs to destroy the old resource and (if required to be present ) re-create it from scratch every time.

File: tests/beaker_tests/snmp_notification_receiver/snmp_notification_receiverlib.rb (hunk 4 above)
Comment (jonnytpuppet) [inline tests/beaker_tests/snmp_notification_receiver/snmp_notification_receiverlib.rb]: There is no concept of default properties as far as this provider is concerned. Items are either set or they aren't. There is the concept of mandatory properties, though.

File: lib/puppet/provider/snmp_notification_receiver/nxapi.rb (hunk 7 above)
Comment (mikewiebe) [inline lib/puppet/provider/snmp_notification_receiver/nxapi.rb]: Does that mean no idempotency?  What if the we discover the resource and all it's properties already exist on the node?

Comment (jonnytpuppet) [inline lib/puppet/provider/snmp_notification_receiver/nxapi.rb]: It is idempotent.

By the time that this line is executed, the prefect has already been run, and the resources are puppet resources.

The flush method only runs if there is a property update.

Perhaps I have misunderstood your question, or I am missing something here. It would be appreciated if you could provide a manifest example or something similar, thanks.

Comment (mikewiebe) [inline lib/puppet/provider/snmp_notification_receiver/nxapi.rb]: No.. you are right.  Still have some 2015 dust on the brain! :)

File: tests/beaker_tests/snmp_notification_receiver/snmp_notification_receiverlib.rb (hunk 4 above)
Comment (mikewiebe) [inline tests/beaker_tests/snmp_notification_receiver/snmp_notification_receiverlib.rb]: Ack.  Please add full test coverage for 1-3.

File: lib/puppet/provider/snmp_notification_receiver/nxapi.rb [hunk 8]
```diff
@@ -0,0 +1,164 @@
...
+    @snmpnotificationreceiver = nil
+
+    return if @property_flush[:ensure] == :absent
+
+    validate
+
+    # Needed due to NXOS issue regarding removing and adding resources too quickly
+    sleep 2
```
mikewiebe: We should not sleep here.  I would prefer you update the node_utils layer delete object method to loop until the resource is gone and raise an error after a reasonable number of checks.  That hides the details from the puppet layer and will be more tolerant of platform differences.

jonnytpuppet: @mikewiebe this pattern was already accepted in snmp_user

https://github.com/cisco/cisco-network-puppet-module/pull/94/files#diff-6e5622cad0a3c8f99872013dbe2dacabR133

I agree it's not ideal, but creates much less overhead than constant re-checking

Thanks

mikewiebe: @jonnytpuppet I did not realize that it was accepted for snmp_user but the issue still remains.  We really should fix this in both places.  In my experience on cisco devices, any time we rely on sleep statements for automation bad things happen where timing can differ across various cisco devices.  We wind up having to change and tweak it later.  Better to write once and build in the flexibility to handle platform differences.

I would loop a predetermined number of times (5-7 for example), sleep 1 second between checks and break out of the loop the second the desired state is reached.  Generate an error if we complete all loops and don't reach the desired state.

jonnytpuppet:
@mikewiebe Code updated, can this be merged please?
The lines lengths are within rubocop limits.

The sleep issues turned out to be moot for this provider, as removing it caused no issues.

Thanks

JT

jonnytpuppet:
tests/beaker_tests/snmp_notification_receiver/snmp_notification_receiver_provider_defaults.rb passed in 85.83 seconds
      Test Suite: tests @ 2016-01-07 16:21:04 +0000

mikewiebe:
:+1:

PR #122, opened 2015-12-24
ipv4_pim_sparse_mode: new property for cisco_interface

chrisvanheuveln:
All of the routedintf tests now pass, except for an issue with the
negative test for channel-group, which correctly hits an error but
for which beaker does not get exit code [1], so I'm temporarily
removing that test.


tests/beaker_tests/cisco_interface/routedintf_provider_negatives.rb
```diff
@@ -510,39 +526,41 @@
...
-      "resource cisco_interface 'ethernet1/4'", options)
-    on(agent, cmd_str) do
-      search_pattern_in_output(stdout,
-                               { 'channel_group' => RoutedIntfLib::CHANNEL_GROUP_NEGATIVE },
-                               true, self, logger)
-    end
-
-    logger.info("Check cisco_interface resource absence on agent :: #{result}")
-  end
+  # TBD: CHANNEL-GROUP NEGATIVE TEST DOES NOT RETURN EXIT CODE [1]
```
chrisvanheuveln: Unrelated bug with channel-group. This is a minor negative test so I'm removing it for now, will create a bug entry for the backlog.

tests/beaker_tests/cisco_interface/routedintflib.rb
```diff
@@ -66,6 +67,17 @@ def self.create_routedintf_manifest_switchport_disabled
       switchport_autostate_exclude => 'default',
       switchport_vtp               => 'default',
       vrf                          => 'default',
+  }}
+EOF"
+    manifest_str
+  end
+
+  def self.create_routedintf_manifest_channel_group
```
chrisvanheuveln: channel-group prevents some L3 properties from being configured so I moved this into it's own test manifest.

glennmatthews:
Instead of `ipv4_pim_sparse_mode` with values `true` and `false`, did you consider `ipv4_pim_mode` with values `sparse`, `dense`, and `sparse-dense`?

lib/puppet/provider/cisco_interface/nxapi.rb
```diff
@@ -55,9 +55,15 @@
     :switchport_trunk_native_vlan,
   ]
   INTF_BOOL_PROPS = [
-    :shutdown, :negotiate_auto, :ipv4_redirects, :ipv4_proxy_arp,
-    :switchport_vtp, :switchport_autostate_exclude,
-    :svi_autostate, :svi_management
+    :shutdown,
```
glennmatthews: If this ordering is necessary here, can you add a comment to that effect? If not, can you put these into alphabetical order?

tests/beaker_tests/cisco_interface/routedintf_provider_nondefaults.rb
```diff
@@ -123,7 +128,7 @@
                                  'switchport_mode'              => 'disabled',
                                  'switchport_vtp'               => 'false',
                                  'vrf'                          => 'test1',
-                                 'channel_group'                => '200' },
```
glennmatthews: Intentional?

chrisvanheuveln:
Meh. :-) Neither NX nor XR support dense mode, but of course there are legacy platforms that do. It's not a huge change but it will require mods to the NU and minitests. I can make the mods if you really want it changed.

PR #123, opened 2016-01-04
add event_history to cisco_bgp

andy91315:
1. pass beaker test
2. pass static analysis

to set no event-history cli I assume
event_history_cli => false
and no value for event_history_cli_size in puppet.
Is that correct?


andy91315:
too many lines while creating the manifest how to fix that?

glennmatthews:
Either refactor the method into one or more shorter methods, or update `.rubocop.yml` with the new higher limit, or wrap the method in `# rubocop:disable Metrics/MethodLength` ... `# rubocop:enable Metrics/MethodLength`.

Looking at the code in question, I'd recommend the third option - disable the metric for this particular method as it's likely to continue to grow over time and I don't see a reasonable way to shorten it.

README.md
```diff
@@ -454,6 +454,37 @@ AS confederation parameters. Valid values are String, keyword 'default'.
 ##### `enforce_first_as`
 Enable/Disable enforces the neighbor autonomous system to be the first AS number listed in the AS path attribute for eBGP. Valid values are 'true', 'false', and 'default'.
 
+##### `event_history_cli`
+Enable/Disable cli event history buffer. Valid values are 'true', 'false', and 'default'.
+
+##### `event_history_cli_size`
+Set cli event history buffer size. Valid values are 'small', 'medium', 'large', 'disable' or 'default'.
+
+##### `event_history_cli`
```
glennmatthews: Looks like `event_history_cli` and `event_history_cli_size` are described twice here.

README.md
```diff
@@ -454,6 +454,37 @@ AS confederation parameters. Valid values are String, keyword 'default'.
 ##### `enforce_first_as`
 Enable/Disable enforces the neighbor autonomous system to be the first AS number listed in the AS path attribute for eBGP. Valid values are 'true', 'false', and 'default'.
 
+##### `event_history_cli`
+Enable/Disable cli event history buffer. Valid values are 'true', 'false', and 'default'.
+
+##### `event_history_cli_size`
+Set cli event history buffer size. Valid values are 'small', 'medium', 'large', 'disable' or 'default'.
```
glennmatthews: What does 'disable' mean in this context? Is it equivalent to setting `event_history_cli` to `false`, or is there a meaningful state where `event_history_cli` is `true` but `event_history_cli_size` is `disable`?

tests/beaker_tests/cisco_bgp/test_bgp_defaults.rb
```diff
@@ -91,6 +91,13 @@
   'bestpath_med_missing_as_worst'          => 'false',
   'bestpath_med_non_deterministic'         => 'false',
   'enforce_first_as'                       => 'true',
+  'event_history_cli'                      => 'true',
+  'event_history_cli_size'                 => 'small',
+  'event_history_detail'                   => 'false',
+  'event_history_events'                   => 'true',
+  'event_history_events_size'              => 'small',
+  'event_history_periodic'                 => 'true',
+  'event_history_periodic_size'            => 'small',
```
glennmatthews: Looks like all of your tests are using either `small` or `large`. Would be good to also exercise `medium` and `disable` values.

README.md
andy91315: it is event_history_cli is true but event_history_cli_size is disable

tests/beaker_tests/cisco_bgp/test_bgp_defaults.rb
andy91315: for test the default the expected value is just small, for non default I have changed to test all values.

README.md
glennmatthews: Sorry, not sure I understand your response. What does it mean as an operational state for the buffer to be enabled (`event_history_cli == true`) but the buffer size is disabled (`event_history_cli_size == disable`)? That doesn't intuitively make any sense to me.

andy91315: But no event history cli and event history cli size disable are two different commands.
first one disable the buffer second one enable the buffer and set the size to 0.

chrisvanheuveln: It's a somewhat confusing cli...   Start from clean system:

```
n31x-101(config)# router bgp 1
n31x-101(config-router)# sh bgp event-history cli 
bgp-1 cli eventsm flag: 0x7, default_size: 32768, configured_size: 0, size: 32768, default_on: Yes, hidden: No

(so it's on by default)

n31x-101(config-router)#   event-history cli size disable
n31x-101(config-router)#   show run bgp
router bgp 1
  event-history cli size disable

n31x-101(config-router)# sh bgp event-history cli 
"event-history cli" is disabled

n31x-101(config-router)# no   event-history cli size disable
n31x-101(config-router)#   show run bgp
router bgp 1
  no event-history cli

(event log is present but no longer updated)
```

glennmatthews: I see that the CLI behaves this way - what is not clear to me is what this actually means. I've searched internally and externally and I can't find anyone actually explaining what the operational difference is between `no event-history cli` and `event-history cli size disable`.

Another way to look at this - these two properties are heavily interdependent, as setting `cli_size` only works if `cli` is true (you can't set `no event-history cli` and `event-history cli size large` simultaneously). Additionally, there's a comment in the CLI suggesting that `event-history cli` is equivalent to `event-history cli size small`, and a quick manual test seems to bear this out:

```
n9k-108(config-router)# no event-history cli 
n9k-108(config-router)# show run bgp all | i cli
  no event-history cli
n9k-108(config-router)# event-history cli 
n9k-108(config-router)# show run bgp all | i cli
  event-history cli
n9k-108(config-router)# event-history cli size small
n9k-108(config-router)# show run bgp all | i cli
  event-history cli
n9k-108(config-router)# event-history cli size medium 
n9k-108(config-router)# show run bgp all | i cli
  event-history cli size medium
n9k-108(config-router)# event-history cli size small
n9k-108(config-router)# show run bgp all | i cli
  event-history cli
```

So it's not clear to me that there's any value in having this be split into two separate properties. You really only have four (*maybe* five) distinct states:

* disabled
* enabled, size small
* enabled, size medium
* enabled, size large
* enabled, size disabled (???)

I think it would really make more sense for this to be a single property.

chrisvanheuveln: During the design phase Andi and I actually discussed doing this as a single property. I don't recall a compelling reason to do it one way or the other at the time, unless it had to do with lack of a size option on some platforms, though I'm not sure XR even supports event-history. If there's no good reason for it then I'm fine with removing the boolean prop.

andy91315: My understanding is that a property can be both boolean and string. how to use one property to differentiate between no event-history cli and event-history cli disabled.
 So I use two properties and refer to default_originate in cisco_neighbor_af.

chrisvanheuveln: @glennmatthews How about something like this:

```
event_history_cli => 'size_disable' 
  # options: true | false | default | size_small | size_medium | size_large | size_disable
  #  true         = event-history cli
  #  false       = no event-history cli
  #  size_small    = event-history cli (small)
  #  size_medium  = event-history cli medium
  #  size_large   = event-history cli large
  #  size_disable = event-history cli size disable
```

glennmatthews: `true` and `size_small` are identical so you'll have to decide which one munges to which other one for idempotence purposes. Other than that this seems fine to me.

andy91315:
Please review this code. Thanks.

PR #124, opened 2016-01-05
Add 'vlan_mapping' attrs to cisco_interface

chrisvanheuveln:
These vxlan Multi-Tenancy Full (MT-full) attrs live under an ethernet interface, e.g.:

```
interface ethernet9/1
  switchport
  switchport mode trunk
  no switchport vlan mapping enable 
  switchport vlan mapping 20 21 
  switchport vlan mapping 30 31 
```

Platform / linecard support is very limited for MT-full (N7k only at the moment but it will later include n5k/n6k) and at the moment there are a lot of pre-requisites to even using this:

- linecard must be type F3
- platform must support VDC
- limit-resource module-type must limit all line cards in the VDC to F3
- switchport mode trunk
- requires bridge-domain global configs

Beaker tests pass on an N7k in default state (F3 card is initially excluded from the VDC).

I added 'SKIP' logic to utililtylib.rb to allow the test to be bypassed if the platform prerequisites cannot be met.

### DIFF
--- a/examples/demo_interface.pp
+++ b/examples/demo_interface.pp
@@ -16,15 +16,16 @@
 
 class ciscopuppet::demo_interface {
   cisco_interface { 'Ethernet1/1' :
-    shutdown            => true,
-    switchport_mode     => disabled,
-    description         => 'managed by puppet',
-    ipv4_address        => '192.168.55.55',
-    ipv4_netmask_length => 24,
-    mtu                 => 1600,
-    speed               => 100,
-    duplex              => 'full',
-    vrf                 => 'test',
+    shutdown               => true,
+    switchport_mode        => disabled,
+    description            => 'managed by puppet',
+    ipv4_address           => '192.168.55.55',
+    ipv4_netmask_length    => 24,
+    ipv4_pim_sparse_mode   => 24,
+    mtu                    => 1600,
+    speed                  => 100,
+    duplex                 => 'full',
+    vrf                    => 'test',
   }
 
   cisco_interface { 'Ethernet1/1.1':
--- a/examples/demo_interface.pp
+++ b/examples/demo_interface.pp
@@ -1,6 +1,6 @@
 # Manifest to demo cisco_interface provider
 #
-# Copyright (c) 2014-2015 Cisco and/or its affiliates.
+# Copyright (c) 2014-2016 Cisco and/or its affiliates.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -56,4 +56,10 @@ class ciscopuppet::demo_interface {
     duplex      => 'auto',
     speed       => '100m',
   }
+  #  Requires F3 or newer linecards
+  # cisco_interface { 'Ethernet9/1':
+  #   switchport_mode                => trunk,
+  #   switchport_vlan_mapping        => [[20, 21], [30, 31]]
+  #   switchport_vlan_mapping_enable => false
+  # }
 }
--- a/tests/beaker_tests/cisco_interface/test_vlan_mapping.rb
+++ b/tests/beaker_tests/cisco_interface/test_vlan_mapping.rb
@@ -1,0 +1,186 @@
+###############################################################################
+# Copyright (c) 2016 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+###############################################################################
+# TestCase Name:
+# -------------
+# test_vlan_mapping.rb
+#
+# TestCase Prerequisites:
+# -----------------------
+# This is a Puppet Interface resource testcase of vlan_mapping properties,
+# for use with Puppet Agent on Nexus devices.
+# The test case assumes the following prerequisites are already satisfied:
+#   - Host configuration file contains agent and master information.
+#   - SSH is enabled on the N9K Agent.
+#   - Puppet master/server is started.
+#   - Puppet agent certificate has been signed on the Puppet master/server.
+#
+###############################################################################
+#
+#        ****************************************
+#        ** IMPORTANT ADDITIONAL PREREQUISITES **
+#        ****************************************
+#
+# The vlan_mapping properties are "Multi-Tenancy Full" properties which
+# currently have limited platform and linecard support. This test script will
+# look for these requirements and fail if they are not present:
+#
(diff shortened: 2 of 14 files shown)


examples/demo_interface.pp
```diff
@@ -56,4 +56,10 @@
     duplex      => 'auto',
     speed       => '100m',
   }
+  #  Requires F3 or newer linecards
```
mikewiebe: Not blocking but eventually I think we will want to use facter facts to set these property values to 'undef' if not supported so that we don't have to leave this commented out.

tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -374,3 +382,131 @@ def bgp_title_pattern_munge(tests, id, provider=nil)
...
+# MT-full currently requires an F3 line module. This method will update
+# the tests hash with the names of the default vdc and also an appropriate
+# test interface name to use for testing.
+# tests[:vdc] The default vdc name
+# tests[:intf] A compatible interface to use for MT-full testing.
+def setup_mt_full_env(tests, testcase)
+  # MT-full tests require a specific linecard. Search for a compatible
+  # module and enable it.
+
+  testheader = tests[:testheader]
```
mikewiebe: Should this be an optional parameter just in case someone does not want to pass this information in?  `testheader = tests[:testheader] unless tests[:testheader].nil?`

tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -374,3 +382,131 @@ def bgp_title_pattern_munge(tests, id, provider=nil)
...
+def prereq_skip(testheader, testcase, message)
+  logger.error("** PLATFORM PREREQUISITE NOT MET: #{message}")
+  raise_skip_exception(testheader, testcase)
+end
+
+# Return an interface name from the first MT-full compatible line module found
+def mt_full_interface
+  # Search for F3 card on device, create an interface name if found
+  cmd = get_vshell_cmd('sh mod')
+  out = on(agent, cmd, pty: true).stdout[/^(\d+)\s.*N7K-F3/]
```
mikewiebe: Would it be better to make this regexp case insensitive?

tests/beaker_tests/cisco_interface/test_vlan_mapping.rb
```diff
@@ -0,0 +1,186 @@
...
+def test_harness_vlan_mapping(tests, id)
+  tests[id][:ensure] = :present if tests[id][:ensure].nil?
+
+  # Build the manifest for this test
+  build_manifest_vlan_mapping(tests, id)
+
+  # Workaround for (ioctl) facter bug on n7k ***
+  tests[id][:code] = [0, 2]
+
+  test_harness_common(tests, id)
```
mikewiebe: We need to remove the `test_show_cmd` from `test_harness_common`

https://github.com/cisco/cisco-network-puppet-module/blob/develop/tests/beaker_tests/lib/utilitylib.rb#L183

Go ahead and take care of that now?

tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -374,3 +382,131 @@ def bgp_title_pattern_munge(tests, id, provider=nil)
...
+  # module and enable it.
+
+  testheader = tests[:testheader]
+  mod = 'f3'
+  step 'Check for Compatible Line Module' do
+    tests[:intf] = mt_full_interface
+    break if tests[:intf]
+    prereq_skip(testheader, testcase,
+                "MT-full tests require #{mod} or compatible line module")
+  end if tests[:intf].nil?
```
mikewiebe: Why is line 404 needed but no similar line is needed for the vdc below?  Once prereq_skip is called we are done right?

tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -374,3 +382,131 @@ def bgp_title_pattern_munge(tests, id, provider=nil)
...
+
+# Add interface to vdc's allocated interfaces
+def vdc_allocate_interface_set(vdc, intf)
+  # Turn off prompting
+  cmd = get_vshell_cmd('terminal dont-ask persist')
+  on(agent, cmd, pty: true)
+
+  cmd = get_vshell_cmd("conf t ; vdc #{vdc} ; "\
+                       "allocate interface #{intf}")
+  on(agent, cmd, pty: true)
```
mikewiebe: Not blocking for this commit but it would be nice to create a wrapper for the on command where we have to configure the device via the vegas shell.  Something like...

`vshell_config("vdc #{vdc}", "allocate interface #{intf}")`

glennmatthews: But our long-term plan is to remove all instances of vshell commands, right?

tests/beaker_tests/lib/utilitylib.rb
chrisvanheuveln: Nope. :-) 

This is the product id from show mod and will always be all caps.

tests/beaker_tests/lib/utilitylib.rb
chrisvanheuveln: Yup, sounds reasonable.

examples/demo_interface.pp
chrisvanheuveln: Agreed.

chrisvanheuveln: Agreed for beaker, but for the demo manifest? 
Won't the agent run fail if the entire provider is unsupported?

tests/beaker_tests/cisco_interface/test_vlan_mapping.rb
chrisvanheuveln: It's out of scope here. We also need to remove every reference for `:show_cmd`, `:show_pattern`, etc.

tests/beaker_tests/lib/utilitylib.rb
chrisvanheuveln: It allows me to completely bypass this `step` if the caller provides `tests[:intf]` instead of having this method figure it out. It's only marginally useful but I found it handy during troubleshooting. 

However, on second look I think it would be useful to add it to the VDC step because that would allow the caller to specify a non-default VDC, so I'll add it and remove the 'default' references from the messages.

tests/beaker_tests/lib/utilitylib.rb
chrisvanheuveln: @mikewiebe The method you're commenting on is already a config wrapper. :-)
Glenn is right this is just a short term solution.

tests/beaker_tests/lib/utilitylib.rb
mikewiebe: ok

tests/beaker_tests/lib/utilitylib.rb
mikewiebe: Well if you are certain it will always be all caps.... :)

tests/beaker_tests/lib/utilitylib.rb
mikewiebe: @glennmatthews Our long term plan is to remove all instances of vshell for validating 'show output'.  Are you also suggesting we remove config cases?  Do you envision we would replace it with similar config api's that we have in minitest?

PR #126, opened 2016-01-07
Beaker test lib facter helper methods

mikewiebe:
New set of simple beaker test lib api's to use `facter` to query for the cisco OS and cisco hardware information.  This is needed to branch test code based on the following.

* Cisco OS Example: `nexus` vs `xr`
* Cisco Hardware Information Example: `n3k`, `n9k`, `n7k` etc...

Note, beaker does have a built in `fact_on` method but I chose not to use this method primarily due to the fact that for custom cisco facts we have to prepend `sudo` to the command and the api does not allow this.

These api's are needed to unblock several teams that are writing test cases for the various platforms, however I am investigating methods to remove the hardcoded path to the `facter` binary (would be useful for `puppet` binary as well).  I would like to handle that in a subsequent PR.

### DIFF
--- a/tests/beaker_tests/lib/utilitylib.rb
+++ b/tests/beaker_tests/lib/utilitylib.rb
@@ -37,6 +37,7 @@
 
 # Define various CONSTANTS used by beaker tests
 PUPPET_BINPATH = '/opt/puppetlabs/bin/puppet '
+FACTER_BINPATH = '/opt/puppetlabs/bin/facter '
 PUPPETMASTER_MANIFESTPATH = '/etc/puppetlabs/code/environments/production/manifests/site.pp'
 
 # These methods are defined outside of a module so that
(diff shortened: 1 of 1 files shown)


tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -510,3 +511,46 @@ def vdc_allocate_interface_set(vdc, intf)
...
+def facter_cmd(cmd)
+  get_namespace_cmd(agent, FACTER_BINPATH + cmd, options)
+end
+
+# Used to cache the operation system information
+@cisco_os = nil
+# Use facter to return cisco operating system information
+def operating_system
+  return @cisco_os unless @cisco_os.nil?
+  @cisco_os = on(agent, facter_cmd('os.name')).stdout.chomp
```
chrisvanheuveln: cleaner:

```
-  return @cisco_os unless @cisco_os.nil?
-  @cisco_os = on(agent, facter_cmd('os.name')).stdout.chomp
+  @cisco_os ||= on(agent, facter_cmd('os.name')).stdout.chomp
```

mikewiebe: Agreed, but maybe less clear?  I know this is a handy trick but I personally prefer the current code even though it's an extra line, I believe the cache intent is more clear.  If you feel very strongly I will change it though... :)

chrisvanheuveln: It's a classic ruby construct. What's not to love? :-)
Okay, not a blocker, I'll merge it.

PR #129, opened 2016-01-11
platform_get module function

mikewiebe:
New puppet module custom function that can be called from a manifest to distinguish between platforms using the cisco custom fact `cisco.hardware.type`.  Currently supports various Nexus platforms but can be extended when needed.

Custom Functions Doc: https://docs.puppetlabs.com/guides/custom_functions.html

NOTE: For NX-OSv the function currently returns 'n9k' but if the team feels like 'nxv'  or something along those lines would be better choice I can make the change.

The function can be called in the manifest like this: `$foo = platform_get()`

### DIFF
--- a/lib/puppet/parser/functions/platform_get.rb
+++ b/lib/puppet/parser/functions/platform_get.rb
@@ -1,0 +1,47 @@
+#
+# Cisco platform_get puppet manifest function.
+#
+# January, 2016 
+#
+# Copyright (c) 2015 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module Puppet::Parser::Functions
+  newfunction(:platform_get, :type => :rvalue) do |args|
+    data = lookupvar('cisco') 
+    pi = data['hardware']['type']
+    # The following kind of string info is returned for Nexus.
+    # - Nexus9000 C9396PX Chassis
+    # - Nexus7000 C7010 (10 Slot) Chassis
+    # - Nexus 6001 Chassis
+    # - NX-OSv Chassis
+    case pi
+    when /Nexus\s?3\d\d\d/
+      cisco_hardware = 'n3k'
+    when /Nexus\s?5\d\d\d/
+      cisco_hardware = 'n5k'
+    when /Nexus\s?6\d\d\d/
+      cisco_hardware = 'n6k'
+    when /Nexus\s?7\d\d\d/
+      cisco_hardware = 'n7k'
+    when /Nexus\s?9\d\d\d/
+      cisco_hardware = 'n9k'
+    when /NX-OSv/
+      cisco_hardware = 'n9k'
+    else
+      raise Puppet::ParseError, "Unrecognized platform type: #{pi}\n"
+    end
+    cisco_hardware
+  end
+end
--- a/lib/puppet/parser/functions/platform_get.rb
+++ b/lib/puppet/parser/functions/platform_get.rb
@@ -32,6 +32,8 @@ module Puppet
         # - Nexus 6001 Chassis
         # - NX-OSv Chassis
         case pi
+        when /Nexus\s?31\d\d/
+          cisco_hardware = 'n31k'
         when /Nexus\s?3\d\d\d/
           cisco_hardware = 'n3k'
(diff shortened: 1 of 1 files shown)


mikewiebe:
Little quick on the draw here.  Rubocop fixes a coming.

mikewiebe:
Rubocop failures resolved

lib/puppet/parser/functions/platform_get.rb
```diff
@@ -0,0 +1,52 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module Puppet
+  module Parser
+    # Function platform_get.  Returns platform string.
+    module Functions
+      newfunction(:platform_get, type: :rvalue) do |_args|
+        data = lookupvar('cisco')
+        pi = data['hardware']['type']
```
glennmatthews: Do we need any logic for failing gracefully if this function is called against a non-Cisco node?

lib/puppet/parser/functions/platform_get.rb
```diff
@@ -0,0 +1,52 @@
...
+        # The following kind of string info is returned for Nexus.
+        # - Nexus9000 C9396PX Chassis
+        # - Nexus7000 C7010 (10 Slot) Chassis
+        # - Nexus 6001 Chassis
+        # - NX-OSv Chassis
+        case pi
+        when /Nexus\s?3\d\d\d/
+          cisco_hardware = 'n3k'
```
glennmatthews: Do we need to distinguish between 30xx vs 31xx?

glennmatthews:
A couple questions but this looks good to me. :+1:

lib/puppet/parser/functions/platform_get.rb
mikewiebe: Not sure.  What differences do you have in mind that we would need to handle?

glennmatthews: I know on the node_utils side there's some code in interface.rb/interface.yaml that specifically checks for 30xx or 31xx. Not sure of the specifics of what it's handling.

lib/puppet/parser/functions/platform_get.rb
```diff
@@ -0,0 +1,54 @@
...
+
+module Puppet
+  module Parser
+    # Function platform_get.  Returns platform string.
+    module Functions
+      newfunction(:platform_get, type: :rvalue) do |_args|
+        data = lookupvar('cisco')
+        fail Puppet::ParseError,
+             "Facter fact 'cisco' unavailable on this platform\n#{__FILE__}" if data.nil?
+        pi = data['hardware']['type']
```
mikewiebe: This now generates the following error when the cisco fact is unavailable:

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Error while evaluating a Function Call, Facter fact 'cisco' unavailable on this platform
/etc/puppetlabs/code/environments/production/modules/ciscopuppet/lib/puppet/parser/functions/platform_get.rb at /etc/puppetlabs/code/environments/production/manifests/site.pp:3:18 on node dt-n9k5-1.cisco.com
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
root@dt-n9k5-1#
```
NOTE: I added `__FILE__` because no path information provided to the offending function.

lib/puppet/parser/functions/platform_get.rb
mikewiebe: Ok, looks like the checks are for speed settings.  Simple enough to add the platform support.

lib/puppet/parser/functions/platform_get.rb
glennmatthews: Would it be better to return `''` or `nil` instead of an error in this case? Just trying to think of the mixed case where the user is managing a mix of Cisco and non-Cisco nodes.

mikewiebe:
@glennmatthews Updated based on your comments.

lib/puppet/parser/functions/platform_get.rb
```diff
@@ -0,0 +1,56 @@
...
+        # The following kind of string info is returned for Nexus.
+        # - Nexus9000 C9396PX Chassis
+        # - Nexus7000 C7010 (10 Slot) Chassis
+        # - Nexus 6001 Chassis
+        # - NX-OSv Chassis
+        case pi
+        when /Nexus\s?31\d\d/
+          cisco_hardware = 'n31k'
```
glennmatthews: 'n31k' suggests 'n31xxx' to me not 'n31xx'. Maybe we should keep this as n3k for both and let the user query the `[hardware][type]` value directly if they need more specificity.

mikewiebe: Yeah... I like the idea of keeping this more general.

lib/puppet/parser/functions/platform_get.rb
mikewiebe: Yeah, that would make it more usable in the manifest for non-cisco cases.  Will update.

glennmatthews:
Come to think of it, a lot of this logic really belongs in node_utils - we need a similar function in node_utils to map product-id to product family. Might be worth a refactor to move most of this into node_utils and avoid duplication of logic.

PR #121, opened 2015-12-23
puppet ace type and provider

yjyongz:
class ciscopuppet::demo_acl {
    cisco_ace { 'ipv4 foo-1 10':
      ensure                 => present,
      action                 => 'permit',
      proto                  => 'tcp',
      src_addr               => '1.2.3.4 2.3.4.5',
      src_port               => 'eq 40',
      dst_addr               => '8.9.0.4/32',
      dst_port               => 'range 32 56',
      option_format          => 'precedence critical',
  }
}

agent-lab3-nx# show run aclmgr

!Command: show running-config aclmgr
!Time: Wed Dec 23 21:59:29 2015

version 7.0(3)I2(1)
ip access-list foo-1
  10 permit tcp 1.2.3.4 2.3.4.5 eq 40 8.9.0.4/32 range 32 56 precedence critical

PR #131, opened 2016-01-11
initial ace commit

yjyongz:
cvh edit: This is a resubmit of an earlier PR (#121).


tests/beaker_tests/cisco_acl/test_ace.rb
```diff
@@ -0,0 +1,222 @@
...
+    'action'        => 'permit',
+    'proto'         => 'tcp',
+    'src_addr'      => '1.2.3.4 2.3.4.5',
+    'src_port'      => 'eq 40',
+    'dst_addr'      => '9.9.0.4/32',
+    'dst_port'      => 'range 32 56',
+  ",
+}
+
+tests['destroy_ipv4_seq_10'] = {
```
chrisvanheuveln: My comment from #121 still needs to be addressed.

https://github.com/cisco/cisco-network-puppet-module/pull/121/files#diff-ac20693befc9cf1b0135cf1ba18823a0R131

tests/beaker_tests/cisco_acl/test_ace.rb
```diff
@@ -0,0 +1,222 @@
...
+  tests[id][:ensure] = nil
+end
+
+#################################################################
+# TEST CASE EXECUTION
+#################################################################
+test_name "TestCase :: #{testheader}" do
+  # -------------------------------------------------------------------
+  logger.info("\n#{'-' * 60}\nSection 1. Default Property Testing")
+  # node_feature_cleanup(agent, 'acl')
```
chrisvanheuveln: Still need to address this #121 comment:
  https://github.com/cisco/cisco-network-puppet-module/pull/121/files#diff-ac20693befc9cf1b0135cf1ba18823a0R204

chrisvanheuveln:
Please ensure that the following are done:

1. beaker tests pass
2. rubocop passes
3. You have a demo manifest in /examples,  ie `examples/demo_acl.pp`

tests/beaker_tests/cisco_acl/test_ace.rb
yjyongz: maybe i didn't get your point, let's set up meeting to close this

chrisvanheuveln: The point I was trying to make is that the ACE provider isn't responsible for creating the ACL itself so I was suggesting that the test should take care of doing that; however, I looked at the ace yaml entry just now and I see that it does a template append which has a side effect of creating the ACL as well. So the second part of my original comment doesn't really apply.

However, you still need to have some kind of cleanup at the beginning of the test. Right now you've got a `# node_feature_cleanup(agent, 'acl')` above. Remove that comment and replace it with a cleanup that removes the `beaker_1` acl from the switch.

PR #134, opened 2016-01-13
acl interface provider/type

bansalpradeep:
rubocop is good. Tested the provider code via example/demo manifests and results were good.
I couldn't run beaker tests because of some environmental issue. 

Wanted to raise a PR while waiting for help on getting the beaker testing done.

thanks
-Pradeep.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   - `channel_group`
   - `ipv4_pim_sparse_mode`
   - `vlan_mapping`, `vlan_mapping_enable`
+  - `ipv4_acl_in`, `ipv4_acl_out`, `ipv6_acl_in`, `ipv6_acl_out`
 - Extended `cisco_vrf` with the following attributes:
   - `vni`
   
--- a/examples/demo_interface.pp
+++ b/examples/demo_interface.pp
@@ -26,6 +26,10 @@ class ciscopuppet::demo_interface {
     speed                  => 100,
     duplex                 => 'full',
     vrf                    => 'test',
+    ipv4_acl_in            => 'foo-1',
+    ipv4_acl_out           => 'foo-2',
+    ipv6_acl_in            => 'foo-3',
+    ipv6_acl_out           => 'foo-4',
   }
 
   cisco_interface { 'Ethernet1/1.1':
--- a/examples/demo_interface.pp
+++ b/examples/demo_interface.pp
@@ -26,10 +26,10 @@ class ciscopuppet::demo_interface {
     speed                  => 100,
     duplex                 => 'full',
     vrf                    => 'test',
-    ipv4_acl_in            => 'foo-1',
-    ipv4_acl_out           => 'foo-2',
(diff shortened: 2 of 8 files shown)


examples/demo_interface.pp
```diff
@@ -26,6 +26,10 @@
     speed                  => 100,
     duplex                 => 'full',
     vrf                    => 'test',
+    ipv4_acl_in            => 'foo-1',
```
mikewiebe: *nit-pick* Prefer we not use `foo`.  Can we pick something more meaningful?

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   - `channel_group`
   - `ipv4_pim_sparse_mode`
   - `vlan_mapping`, `vlan_mapping_enable`
+  - `ipv4_acl_in`, `ipv4_acl_out`, `ipv6_acl_in`, `ipv6_acl_out`
```
mikewiebe: Please make sure to update the resource reference for these properties.

https://github.com/cisco/cisco-network-puppet-module/blob/develop/README.md#type-cisco_interface

tests/beaker_tests/cisco_interface/routedintf_provider_negatives.rb
```diff
@@ -526,6 +526,174 @@
     logger.info("Check interface instance absence on agent :: #{result}")
   end
 
+  # @step [Step] Requests manifest from the master server to the agent.
```
mikewiebe: Did you mean to add these tests to routedintf_provider_negatives.rb.  The tests are broken down into 3 categories (default, non-default, negatives) 

default = set all manifest values to default and test
non_default = set all manifest values to non-default and test
negative = set manifest values to invalid input and test.

tests/beaker_tests/cisco_interface/routedintf_provider_nondefaults.rb
```diff
@@ -152,6 +156,10 @@
                                  /speed 100/,
                                  /duplex full/,
                                  /vrf member test1/,
+                                 /ip access-group v4acl1 in/,
```
mikewiebe: Don't worry about validating show output.. just use the `puppet resource` command to validate changes.

tests/beaker_tests/cisco_interface/routedintf_provider_negatives.rb
bansalpradeep: Yes, I am following the existing "negative" cases and they are coded the same way. "defaults" and "non-defaults" are created by loading manifests consisting of all configs.

tests/beaker_tests/cisco_interface/routedintf_provider_nondefaults.rb
bansalpradeep: I will just leave it there, since it is already coded up. what say?

mikewiebe: We need to remove them.  This will not scale across multiple platforms but using the `puppet resource` command can be used on any platform that supports puppet.

PR #140, opened 2016-01-20
Add manifest functions: find_linecard, prop_supported

chrisvanheuveln:
-Fix demo_vrf.pp to use these functions

### DIFF
--- a/lib/puppet/parser/functions/find_linecard.rb
+++ b/lib/puppet/parser/functions/find_linecard.rb
@@ -1,0 +1,40 @@
+#
+# Cisco find_linecard puppet manifest function.
+#
+# January 2016, Chris Van Heuveln
+#
+# Copyright (c) 2016 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


lib/puppet/parser/functions/find_linecard.rb
```diff
@@ -0,0 +1,40 @@
...
+    # Input : String, Linecard name, eg. 'N7K-F3'
+    # Output: Slot number of first match found
+    #
+    module Functions
+      newfunction(:find_linecard, type: :rvalue) do |args|
+        data = lookupvar('cisco')
+        return '' if data.nil?
+
+        pat = Regexp.new(args[0])
+        inv = data['inventory']
```
mikewiebe: Do you  also need a nil check for `data['inventory']`?

lib/puppet/parser/functions/find_linecard.rb
```diff
@@ -0,0 +1,40 @@
...
+        match = inv.keys.select { |slot| inv[slot]['pid'].match(pat) }
```
mikewiebe: How are multiple matches treated here.  Does it just return the first match?

lib/puppet/parser/functions/find_linecard.rb
chrisvanheuveln: Nope. `inv.keys` is an array and it's enumerable.

lib/puppet/parser/functions/find_linecard.rb
chrisvanheuveln: First match. See comment on line 26.

mikewiebe:
:+1:

PR #141, opened 2016-01-22
Update utilitylib.rb

LuvCurves:
Beaker now supports getting VRF from host object. Requires Beaker version 2.33.0

### DIFF
--- a/tests/beaker_tests/lib/utilitylib.rb
+++ b/tests/beaker_tests/lib/utilitylib.rb
@@ -51,13 +51,7 @@ PUPPETMASTER_MANIFESTPATH = '/etc/puppetlabs/code/environments/production/manife
 # @result namespacestr [String] Returns 'sudo ip netns exec vrf <cmd>'
 # command string for 'cisco' platform.
 def get_namespace_cmd(host, cmdstr, options)
-  case host['platform']
-  when /cisco/
-    agentvrf = options[:HOSTS][host.to_s.to_sym]['vrf']
-    return "sudo ip netns exec #{agentvrf} " + cmdstr
-  else
-    return cmdstr
-  end
+  return cmdstr
 end
 
 # Method to return the Vegas shell command string for a NXOS CLI command.


LuvCurves:
DO NOT MERGE UNTIl BEAKER 2.33.0 is released.

tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -51,13 +51,7 @@
...
 # command string for 'cisco' platform.
 def get_namespace_cmd(host, cmdstr, options)
-  case host['platform']
-  when /cisco/
-    agentvrf = options[:HOSTS][host.to_s.to_sym]['vrf']
-    return "sudo ip netns exec #{agentvrf} " + cmdstr
-  else
-    return cmdstr
-  end
+  return cmdstr
```
chrisvanheuveln: Your PR description says: "Beaker now supports getting VRF from host object."

Your change no longer references host. This method is supposed to return cmdstr with the appropriate namespace / vrf context.

So I don't quite understand. Can you explain how this is supposed to work in a vrf after your change?

LuvCurves: Using the vrf context will all be handled by beaker.  Beaker will store the value of the vrf and use it when executing commands on the switch.  There really won't even be a need for this function once the next version of Beaker is released.

chrisvanheuveln: Ah, okay. Can you point me at the beaker code where these vrf changes are being made? Thanks.

glennmatthews:
beaker 2.33 (and 2.34) are now released. What can we do to move this forward?

tests/beaker_tests/lib/utilitylib.rb
LuvCurves: You can look here for the changes.  https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/cisco.rb

It was released last week

LuvCurves:
@glennmatthews I've added a comment in-line.  I previously did not have the final changes to beaker I could point @chrisvanheuveln at but now you can go here to see them: https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/cisco.rb

chrisvanheuveln:
@LuvCurves Can you remove the `return` to make rubocop happy? :+1: after that.

@glennmatthews I ran a quick test with these diffs and beaker 2.34.0. Looks good except for the rubocop failure.

LuvCurves:
@chrisvanheuveln Mistakenly dropped my forked repo.  Going to close this PR and open a new one that passes rubocop.

PR #150, opened 2016-01-26
Convert cisco_interface beaker tests to new hash syntax

chrisvanheuveln:
I combined the tests into one test file.

- Still an outstanding issue inre: speed/dup/mtu handling with plat/linecard differences

- In the interest of a small diff, I haven't removed the old test files yet. I'll remove those separately after this merges.


mikewiebe:
Which platforms has this test been run on?

chrisvanheuveln:
@mikewiebe N7k only. As discussed earlier this was a conversion effort only. We still need to address the speed/dup/mtu issue and fully test on all platforms.

tests/beaker_tests/cisco_interface/test_interface.rb
```diff
@@ -0,0 +1,465 @@
...
+end
+
+#################################################################
+# TEST CASE EXECUTION
+#################################################################
+test_name "TestCase :: #{testheader}" do
+  # -------------------------------------------------------------------
+  logger.info("\n#{'-' * 60}\nSection 1. (L3) Property Testing")
+
+  tests[:title_pattern] =
```
robert-w-gries: Remove this line

PR #154, opened 2016-01-27
fix bgp test file for n7k

andy91315:
1. cisco_bgp/test_title_patten
    cisco_bgp_neighbor/test_non_default.rb
    has fails  unable to fix them, logic seems correct to me.

2. delete all show command "show running-config section bgp "in cisco_bgp test
3. substitute node_clean_agent
4. differentiate properties for different platform assuming n5k|n6k|n7k   n3k|n9k  two groups


tests/beaker_tests/cisco_bgp/bgplib.rb
```diff
@@ -237,10 +234,28 @@ def self.create_bgp_manifest_present_non_default
       }
     }
     EOF"
+
     manifest_str
   end
   # rubocop:enable Metrics/MethodLength
 
+  # creat manifest for difference on platform
```
glennmatthews: Can this be expanded on for clarity? What difference(s), what platform(s), why is this needed?

tests/beaker_tests/cisco_bgp/test_bgp_defaults.rb
```diff
@@ -135,6 +135,12 @@
   # Default VRF Test Cases
   # ----------------------
 
+  # delete properties not support on platform
+  device = platform
```
glennmatthews: Perhaps out of scope for this PR but this is going to be painful to merge into feature/grpc as there we are using 'platform' to mean Nexus vs XR, not N7k vs N3k. `platform` as used in this context should really be `product_id` or `product_grouping`.

tests/beaker_tests/cisco_bgp/test_bgp_non_defaults.rb
```diff
@@ -116,6 +113,13 @@
   'timer_bgp_holdtime'                     => '110',
 }
 
+expected_values_diff = {
```
glennmatthews: Instead of this approach of creating an entirely separate hash, consider modifying the contents of `expected_values` based on the platform. This is what we're already doing in feature/grpc for Nexus/XR differences - see https://github.com/cisco/cisco-network-puppet-module/blob/feature/grpc/tests/beaker_tests/bgp/test_bgp_non_defaults.rb#L101 for examples. If you can follow this approach now it will make the subsequent merge a lot less painful.

andy91315: How to build different in bgplib for the manifest? It is built as a string and add to file manifest.pp.

glennmatthews: Similar approach - make the platform a parameter to the method, and define part of the string depending on the platform:

https://github.com/cisco/cisco-network-puppet-module/blob/feature/grpc/tests/beaker_tests/bgp/bgplib.rb#L37
https://github.com/cisco/cisco-network-puppet-module/blob/feature/grpc/tests/beaker_tests/bgp/bgplib.rb#L40
https://github.com/cisco/cisco-network-puppet-module/blob/feature/grpc/tests/beaker_tests/bgp/bgplib.rb#L86

tests/beaker_tests/cisco_bgp/test_bgp_defaults.rb
mikewiebe: @glennmatthews Agreed, out of scope here. I can make the change to re-name the `platform` api to `product_id` and fix up any tests using it.

mikewiebe:
@andy91315 I propose that we move this testing into the new Beaker template instead of attempting to make such drastic changes to the old template for cross-Nexus platform support.  @glennmatthews , @chrisvanheuveln if you guys agree, let's close this and I can start on that tomorrow.  This is the only bgp test that uses the old template and we need to fix that for release 1.2.0

glennmatthews:
It'll be painful to merge with the gRPC code but I agree it's the right direction to move forward with.

chrisvanheuveln:
@mikewiebe  I agree. It's a little more cost now but worth it. With all the helpers we've added recently that will hopefully make the conversion go quicker.

mikewiebe:
@glennmatthews @chrisvanheuveln 
- Ok we will execute on the plan to move beaker tests to the new template. 

@andy91315 
- Please remove the beaker files from the diffs but leave the demo manifest diffs.

PR #157, opened 2016-01-28
Tighten up testbed setups for vdc, vlan_mapping, service_vni

chrisvanheuveln:
This started off as a fix for bridge-domain but morphed into additional cleanups.

1. bridge-domain setup will fail if there's currently a bridge-domain config present on the testbed; so I had to check the current state and clean it up before adding the domain for the test.
2. This led to creating a specific config_bridge_domain helper method, which led to similar helper methods for the other mt_full_env setups. This area could probably still use some work but my commit gets it closer.

### DIFF
--- a/tests/beaker_tests/cisco_interface/test_vlan_mapping.rb
+++ b/tests/beaker_tests/cisco_interface/test_vlan_mapping.rb
@@ -83,15 +83,15 @@ testheader = 'Resource cisco_interface: vlan_mapping properties'
 # Top-level keys set by caller:
 # tests[:master] - the master object
 # tests[:agent] - the agent object
-# tests[:config_bridge_domain] - the bridge-domain configuration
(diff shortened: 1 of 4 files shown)


tests/beaker_tests/cisco_interface/test_vlan_mapping.rb
```diff
@@ -83,15 +83,15 @@
 # Top-level keys set by caller:
 # tests[:master] - the master object
 # tests[:agent] - the agent object
-# tests[:config_bridge_domain] - the bridge-domain configuration
-# tests[:config_switchport] - the interface switchport configuration
+# tests[:bridge_domain] - the bridge-domain configuration
+# tests[:switchport] - the interface switchport mode type
```
robert-w-gries: s/switchport/switchport_mode

robert-w-gries:
:+1:

PR #158, opened 2016-01-28
add demo_bgp.pp

andy91315:
add demo_bgp.pp only


lib/puppet/provider/cisco_ace/nxapi.rb
```diff
@@ -1,9 +1,9 @@
 #
```
mikewiebe: Please remove these cisco_ace diffs from your diffs.

PR #162, opened 2016-01-29
Demo Interface manifest fixes

robert-w-gries:
### Fixes

* Create acl dependencies before setting `ip_acl` properties
* Removed `speed` and `duplex` from demo manifest due to too many differences between platforms and linecards
* Use different interfaces for `channel_group` and `switchport mode access` testing

### Testing

* 3k - passes
* 5k - **Needs testing**
* 6k - **Needs testing**
* 7k - passes
* 9k - passes

### DIFF
--- a/examples/demo_interface.pp
+++ b/examples/demo_interface.pp
@@ -15,6 +15,26 @@
 # limitations under the License.
 
 class ciscopuppet::demo_interface {
+  cisco_acl { 'ipv4 v4acl1':
+    before => Cisco_interface['Ethernet1/1'],
+    ensure => 'present',
+  }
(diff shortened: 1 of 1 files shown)


examples/demo_interface.pp
```diff
@@ -42,14 +63,14 @@
     channel_group   => 200,
```
chrisvanheuveln: You don't appear to be on the latest refpoint. The channel_group property was moved out of cisco_interface already:

https://github.com/cisco/cisco-network-puppet-module/blob/release_1.2.0/examples/demo_interface.pp#L41

PR #163, opened 2016-01-29
Use TPNNS SSHd. Don't hard-code user/pass in utilitylib.

glennmatthews:
Update the workflow in the Beaker readme to explain setting up TPNNS SSHd and non-root login on XR. Update utilitylib to use the SSH user/password from hosts.cfg instead of hard-coding 'root/lab'.

### DIFF
--- a/docs/README-develop-beaker-scripts.md
+++ b/docs/README-develop-beaker-scripts.md
@@ -63,9 +63,9 @@ end
 
 ### Configure IOS XR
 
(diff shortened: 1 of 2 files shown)


glennmatthews:
@sushirol @chburket @brturn Have any of you had a chance to try this diffset out? It works for me and I think makes for a saner workflow but I'd like some feedback before I merge. Thanks!

glennmatthews:
New diffs with hopefully improved IOS XR configuration steps. @chburket please take another look. Thanks!

chburket:
:+1:

PR #175, opened 2016-02-04
Copy relevant git hooks from cisco-network-node-utils

glennmatthews:
Copying over some of the most useful git hooks from cisco-network-node-utils. Others can be ported over as needed.


robert-w-gries:
:+1:

Suggestion:  adding a puppet-lint check for the demo manifests could be useful

PR #176, opened 2016-02-04
beaker testcase rewrites for utilitylib, bgp, pim

chrisvanheuveln:
## beaker testcase rewrites for utilitylib, bgp, pim

This commit suffers from some serious code creep. I had intended to simply convert the cisco bgp test files to use the new hash format but I kept find redundancies and one change led to another until... this.

* README-beaker-script-ref.md: I got tired of all the inaccurate cut & paste comments (on test variables, etc) from each test file so I just removed all the comments and wrote this doc as a single reference. It is currently focused on test variables and the tests hash but I think it should eventually be combined with the earlier README-develop-beaker-scripts.md doc. I'll tackle it pre-release if we have time.

* Most of the `tests[id][:resource]` entries were not needed after setting up the tests to use prop_hash_to_manifest()

* I created common methods for creating manifests, resource cmd result hashes, and title pattern munging

* I rewrote the pim and vrf tests because they used their own pattern munging methods and I wanted them to use the new common munger

* All of the tests I changed pass 100% on an N9k. I still need to validate on N3/5/6/7 but at this point any problems will be small follow-up commits


glennmatthews:
Oof, you aren't making it easy for the next feature/grpc merge. :-)

tests/beaker_tests/README-beaker-script-ref.md
```diff
@@ -0,0 +1,232 @@
+# Beaker Test Script Guide to Test Inputs
```
glennmatthews: 'tests/beaker_tests/README-beaker-script-ref.md'

1) Does 'ref' mean 'reference', 'refactor', or what?
2) Why in tests/beaker_tests/ instead of in doc/? I'd prefer to keep all documentation in doc/.

tests/beaker_tests/README-beaker-script-ref.md
```diff
@@ -0,0 +1,232 @@
...
+
+#### Puppet Command Return Codes
+
+* `0` = No Changes have occurred (idempotence tests)
+* `1` = An Error has been raised
+* `2` = Changes have occurred (manifest tests)
+* `4` = A Failure has occurred
+* `6` = A Change and Failure has occurred
+
+*Note: The `on()` method returns `0` as its default exit code*
```
glennmatthews: Not clear to me what this means - doesn't it return the exit code of whatever command was run? If 0 is the default, how do I override the default?

tests/beaker_tests/README-beaker-script-ref.md
```diff
@@ -0,0 +1,232 @@
...
+
+Custom keys may be added for various unique test requirements but they should be documented in this guide.
+
+### <a name="tlk">Top-level Keys</a>
+
+*Req = Required, Opt = Optional*
+
+|     | Key Name         | Value and Usage
+|:--- |:-----------------|:---------------
+| Req | `:agent`         | The `agent` object variable name. It is used for accessing the agent environment.
```
glennmatthews: Might be more readable to split this into two tables under subheadings '#### Required Keys' and '#### Optional Keys'

tests/beaker_tests/README-beaker-script-ref.md
```diff
@@ -0,0 +1,232 @@
...
+  }
+}
+```
+
+This simple test case definition tells the test harness to:
+
+* set `:title_pattern` and `:desc` to key name (`'basic_1'`)
+* ignore any platform checks (`:platform` is not specified)
+* skip pre cleaning (`:preclean` is not specified)
+* create a `:resource` expected values hash from `:manifest_props` since the expect values from `puppet resource` will be the same as those in the manifest
```
chburket: ...since the expect**ed** values...

tests/beaker_tests/README-beaker-script-ref.md
```diff
@@ -0,0 +1,232 @@
...
+| Opt | `:encap_prof_global` | The global encapsulation profile configuration. This configuration is a dependency for testing with `cisco_interface_service_vni`.
+| Opt | `:intf_type`     | The interface type string, e.g. `ethernet`. Used in tests that require an interface, in which case the test will discover the first interface of that type to use for the tests.
+| Opt | `:platform`      | A regexp pattern to match against the agent's product-id value. This is used to skip tests that don't support a feature or parameter. This key can be specified as a common top-level key or on a test-by-test basis.
+| Opt | `:preclean`      | The resource name used to remove a configuration as part of testbed pre-cleanup. The resource does not have to be the same as the one being tested; for example, the `'cisco_bgp_af'` test might use `'cisco_bgp'` for `:preclean` if a full bgp cleanup is needed instead of just removing the `'cisco_bgp_af'` configuration. This key can be specified as a common top-level key or on a test-by-test basis.
+| Opt | `:sid`           | The service ID (Integer). Used in `cisco_interface_service_vni` tests.
+
+##### Example: Top-level Keys
+
+```
```
glennmatthews: Can you mark these various code blocks as Ruby (```` ```ruby````)?

tests/beaker_tests/README-beaker-script-ref.md
mikewiebe: The default `:acceptable_exit_code` for the `on()` method is `0`.  An error will be thrown if the exit code does not match.  This can be customized using `:acceptable_exit_codes` in the options hash.

glennmatthews: That would be much clearer - thanks Mike.

tests/beaker_tests/README-beaker-script-ref.md
```diff
@@ -0,0 +1,232 @@
...
+
+### <a name="tlk">Top-level Keys</a>
+
+*Req = Required, Opt = Optional*
+
+|     | Key Name         | Value and Usage
+|:--- |:-----------------|:---------------
+| Req | `:agent`         | The `agent` object variable name. It is used for accessing the agent environment.
+| Req | `:master`        | The `master` object variable name. It is used for accessing the puppet master environment.
+| Req | `:resource_name` | The resource name used in the test. This is used with the `puppet resource` command as well as the testheader value in some tests.
```
mikewiebe: *nit* You did not define `testheader`.

tests/beaker_tests/cisco_bgp/test_bgp.rb
```diff
@@ -0,0 +1,260 @@
...
+    'timer_bgp_holdtime'                     => '180',
+    'timer_bgp_keepalive'                    => '60',
+  },
+}
+
+tests[:default_plat_1] = {
+  desc:           '1.2 Default Properties Platform-specific Part 1',
+  platform:       'n(3|9)k',
+  title_pattern:  "#{asn} default",
+  code:           [0, 2],
```
mikewiebe: If we are cleaning before the test then shouldn't this just be looking for 2?  Also, comes to mind that it's not clear from the docs or key name what this is referring to.  `puppet resource` command or something else?

tests/beaker_tests/cisco_bgp/test_bgp.rb
```diff
@@ -0,0 +1,260 @@
...
+    'shutdown'                               => 'false',
+    'suppress_fib_pending'                   => 'false',
+    'timer_bestpath_limit'                   => '300',
+    'timer_bestpath_limit_always'            => 'false',
+    'timer_bgp_holdtime'                     => '180',
+    'timer_bgp_keepalive'                    => '60',
+  },
+}
+
+tests[:default_plat_1] = {
```
mikewiebe: Move this under the platform specific section below?

tests/beaker_tests/cisco_bgp/test_bgp.rb
```diff
@@ -0,0 +1,260 @@
...
+  resource:       { 'ensure' => 'present' },
+}
+
+# Title Pattern Test Hash
+titles = {}
+titles['T.1'] = {
+  title_pattern: '2',
+  title_params:  { vrf: 'blue' },
+}
+
```
mikewiebe: Also need the following test:

```
titles['T.2'] = {
  title_pattern: 'new_york',
  title_pararms: { asn: '55.5' , 'red' },
}
```

tests/beaker_tests/cisco_bgp/test_bgp.rb
```diff
@@ -0,0 +1,260 @@
...
+#  - command return codes
+#  - A description of the 'tests' hash and its usage
+#
+###############################################################################
+require File.expand_path('../../lib/utilitylib.rb', __FILE__)
+
+testheader = 'Resource cisco_bgp'
+
+# Test hash top-level keys
+asn = '1'
```
mikewiebe: Add dotplus notation testing?

tests/beaker_tests/README-beaker-script-ref.md
chrisvanheuveln: 1) I was trying to keep the filename short - it was intended to mean 'reference'. 
2) I thought about putting it in doc/ but liked the handiness of having the yaml doc in cmd_ref/, so I kept this one in beaker_tests/.

In the long run this document needs to be part of README-develop-beaker-scripts.md anyway, so I'll get rid of this document and add this content as an addendum to the main doc.

tests/beaker_tests/cisco_bgp/test_bgp.rb
```diff
@@ -0,0 +1,260 @@
...
+  logger.info("\n#{'-' * 60}\nSection 1. Default Property Testing")
+
+  # -----------------------------------
+  test_harness_run(tests, :default)
+  test_harness_run(tests, :default_plat_1)
+
+  id = :default
+  tests[id][:ensure] = :absent
+  tests[id].delete(:preclean)
+  test_harness_run(tests, id)
```
mikewiebe: Why are you running the `:default` testcase twice?  Once on line 221 and again on line 227?

tests/beaker_tests/cisco_bgp/test_bgp.rb
mikewiebe: Looks like this is needed for the other bgp tests as well.

tests/beaker_tests/cisco_bgp_neighbor_af/test_bgpneighboraf.rb
```diff
@@ -143,550 +70,242 @@
...
-  # Special Cases
-  test_harness_bgp_nbr_af(tests, 'non_default_properties_ibgp_only_l2vpn')
-  test_harness_bgp_nbr_af(tests, 'non_default_misc_maps_part_1_l2vpn')
+  logger.info("\n#{'-' * 60}\nSection 3. L2VPN Property Testing")
+  resource_absent_cleanup(agent, 'cisco_bgp', 'BGP CLEAN :: ')
+  title = '2 default 1.1.1.1 l2vpn evpn'
+  [
+    # Commented test cases are not supported by l2vpn/evpn
+    # :non_def_A2,
+    # :non_def_A3,
```
mikewiebe: Why not just remove these?

tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -455,87 +456,159 @@ def bgp_nbr_remote_as(agent, remote_as)
...
 
-  t[:afi], t[:vrf], t[:rp_addr], t[:group] = title.split
-  t.merge!(tests[id][:pim_group])
-  t[:vrf] = 'default' if t[:vrf].nil?
-  t
+  resource_absent_cleanup(agent, tests[id][:preclean]) if
+    tests[id][:preclean]
+
+  # Set up remote-as if necessary
+  bgp_nbr_remote_as(agent, tests[id][:remote_as]) if tests[id][:remote_as]
```
mikewiebe: Seems odd to have this in a general purpose method.  Should we extract this out into another method called `process_custom_keys` or something along those lines?

tests/beaker_tests/README-beaker-script-ref.md
chrisvanheuveln: I fixed the wording.

tests/beaker_tests/README-beaker-script-ref.md
chrisvanheuveln: I had the same thought after I posted. Already changed it.

tests/beaker_tests/README-beaker-script-ref.md
chrisvanheuveln: Yeah, I'd like to just get rid of it altogether. I don't use it in the new test files but some of the utilitylib methods still have it as a local variable. More:

* `testheader` is a standalone variable in the old tests
* I had a :testheader key in some of the tests to use with the skip logic but it's really just a log message and seems unnecessary
* After I created `:resource_name` I realized I could overload its use in the skip logic (`testheader = 'Resource cisco_foo'`, while `tests[:resource_name] = 'cisco_foo'` so I yanked it out of the new test files.

tests/beaker_tests/README-beaker-script-ref.md
chrisvanheuveln: done

tests/beaker_tests/README-beaker-script-ref.md
glennmatthews: FWIW I've got a lot of regrets about putting the YAML readme in cmd_ref/ instead of doc/, so let's not try to continue that anti-pattern. :-)

tests/beaker_tests/cisco_bgp/test_bgp.rb
chrisvanheuveln: Hmm. I thought I had added that earlier already. 

It's there now.

glennmatthews:
Any updates to the test template files for your changes?

tests/beaker_tests/README-beaker-script-ref.md
chrisvanheuveln: Agreed. It's out of place but it *is* handy. :-)

tests/beaker_tests/cisco_bgp/test_bgp.rb
chrisvanheuveln: I added it to the title pattern test cases.

tests/beaker_tests/cisco_bgp/test_bgp.rb
chrisvanheuveln: I put it here because I'm doing a 'default' test. It would be in the regular default test case if it were supported on all platforms.

tests/beaker_tests/cisco_bgp_neighbor_af/test_bgpneighboraf.rb
chrisvanheuveln: I meant to have some indication that these are explicitly skipped. But I'm fine with removing them.

tests/beaker_tests/lib/utilitylib.rb
chrisvanheuveln: I agree. These dependencies are bound to grow so better to move them into their own method.

tests/beaker_tests/cisco_bgp/test_bgp.rb
chrisvanheuveln: I removed it. I just put the test in a different vrf so that it still passes the manifest test (and doesn't require another preclean).

chrisvanheuveln:
@glennmatthews *"Any updates to the test template files for your changes?"*

The whole beaker dev doc needs to be updated. I'll take a crack at that after this code gets merged.

chrisvanheuveln:
:shipit: ?

docs/README-develop-beaker-scripts.md
```diff
@@ -866,3 +867,259 @@ Inspecting 2 files
...
+* `1` = An Error has been raised
+* `2` = Changes have occurred (manifest tests)
+* `4` = A Failure has occurred
+* `6` = A Change and Failure has occurred
+
+*Note: The beaker `on()` method has an optional `options` parameter, which allows the caller to specify `:acceptable_exit_codes`. If this is not specified it will default to `0`.*
+
+## <a name="tests">The `tests` hash</a>
+
+The tests hash is a convenient container for common and custom test variables. Common variables that are used by all test cases are defined as top-level keys, while each test case is a top-level key with the test case variables defined as the key-value; for example:
```
glennmatthews: The `tests` hash...

tests/beaker_tests/cisco_bgp/test_bgp.rb
```diff
@@ -0,0 +1,267 @@
...
+    timer_bestpath_limit:        '255',
+    timer_bestpath_limit_always: 'true',
+    timer_bgp_holdtime:          '110',
+    timer_bgp_keepalive:         '45',
+  },
+}
+
+# Platform-specific tests
+tests[:non_def_plat_1] = {
+  desc:           '3.1 Default Properties Platform-specific Part 1',
```
glennmatthews: FWIW, the approach we've been going with in XR to date has been different - rather than this approach, we've been modifying a shared test hash depending on the platform (so this would be a customization to `:test_non_def_4` rather than a separate test case). Thoughts as to the pros and cons of this approach versus that one? Does this provide more 'auto-magic'?

tests/beaker_tests/cisco_bgp/test_bgp.rb
```diff
@@ -0,0 +1,267 @@
...
+    fast_external_fallover:                 'true',
+    flush_routes:                           'false',
+    graceful_restart:                       'true',
+    graceful_restart_helper:                'true',
+    graceful_restart_timers_restart:        '130',
+    graceful_restart_timers_stalepath_time: '310',
+  },
+}
+
+tests[:non_def_4] = {
```
glennmatthews: What's the benefit or reasoning behind having four separate test cases rather than a single test case (or for that matter one test case per property or property set)? Do we have guidelines for this?

tests/beaker_tests/cisco_bgp/test_bgp.rb
```diff
@@ -0,0 +1,267 @@
...
+
+tests[:title_patterns_2] = {
+  desc:          'T.2 Title Pattern',
+  title_pattern: '11.4',
+  title_params:  { vrf: 'blue' },
+  resource:      { 'ensure' => 'present' },
+}
+
+# This helper tests a test case in vrf context. This allows for testing a vrf
+# while an existing config is present in vrf default.
```
glennmatthews: This is going to be tricky to adapt to XR - XR has a *lot* of properties that are not configurable under VRFs.

mikewiebe:
:+1:

glennmatthews:
:+1: with some reservations - I know 1.2.0 is under time pressure but this will probably need to be revisited for XR support.

docs/README-develop-beaker-scripts.md
chrisvanheuveln: fixed

tests/beaker_tests/cisco_bgp/test_bgp.rb
chrisvanheuveln: I had tried adding some platform logic to the test hash definitions but I don't recall now why it didn't work and why I abandoned it. If you've got it working then I'd prefer to use that where possible; the fewer number of test cases the better.

tests/beaker_tests/cisco_bgp/test_bgp.rb
chrisvanheuveln: I included this because the old tests had coverage for it but I'm not convinced that these extra vrf tests are really needed. This just adds a different context to the previous tests. Plus we've already got vrf testing in minitest. I'd be happy to remove them.

glennmatthews: We need to cover VRF in beaker too. Maybe not 100% parallel coverage but at least we need to make sure that the type basically works in a VRF. (Aside: it would have been a lot easier for XR to have double the number of types - cisco_bgp/cisco_bgp_vrf, cisco_bgp_af/cisco_bgp_vrf_af, cisco_bgp_neighbor/cisco_bgp_vrf_neighbor, cisco_bgp_neighbor_af/cisco_bgp_vrf_neighbor_af - but oh well. :-) )

tests/beaker_tests/cisco_bgp/test_bgp.rb
chrisvanheuveln: It's a holdover from the first beaker rewrite effort. At the time it was a hassle to troubleshoot 30 properties so breaking them up into smaller chunks made sense. Now that the test hash entries are more concise (:resource is automatically generated) there's not a lot of reason to keep them separate. I can't change these now (no more refactoring allowed!) but I can do it post-1.2.0, or let XR take this on.

PR #181, opened 2016-02-08
Update demo_bgp.pp

HelenHZhang:
update bgp_neighbor_af for IOS-XR

### DIFF
--- a/examples/demo_bgp.pp
+++ b/examples/demo_bgp.pp
@@ -333,89 +333,75 @@ class ciscopuppet::demo_bgp {
   }
 
   # --------------------------------------------------------------------------#
-  # Configure Neighbor-level Address Family IPv4 Unicast
+  # Configure Neighbor-level Address Family IPv4 Unicast (default vrf)
   # --------------------------------------------------------------------------#
+  cisco_bgp_af { '55.77 default ipv4 unicast':
+    ensure                                 => present,
+  }
+  cisco_bgp_neighbor { '55.77 default 1.1.1.1':
+    ensure                                 => present,
+    remote_as                              => 2,
+  }
+
+  cisco_bgp_neighbor_af { '55.77 default ipv4 unicast':
+    ensure                      => present,
+
+    # Properties
+    allows_in                   => 'default',
+    allowas_in_max              => 5,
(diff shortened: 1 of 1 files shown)


examples/demo_bgp.pp
```diff
@@ -333,89 +333,75 @@
...
-    unsuppress_map              => 'unsup_map',
     weight                      => 30,
   }
-
-  cisco_bgp_neighbor_af { '55.77 blue 2.2.2.2 ipv4 unicast':
+ 
+  # --------------------------------------------------------------------------#
+  # Configure Neighbor-level Address Family IPv4 Unicast (non-default vrf)
+  # --------------------------------------------------------------------------#
+  cisco_bgp { '55.77':
```
chburket: Not a big deal, but maybe you could make use of some of the bgp / bgp_af / bgp_neighbor config farther up, instead of adding so much dependency config here?  Add attributes to the earlier config, if you need to (like router_id, etc).

chburket:
And did you make sure this works on both xr and nx?

examples/demo_bgp.pp
glennmatthews: Agreed. Let's keep this manifest as simple as possible.

HelenHZhang:
updated for both platforms. "puppet agent -t" on both platform passed for default-vrf and non-default-vrf.

chburket:
:+1:

examples/demo_bgp.pp
```diff
@@ -333,89 +333,78 @@
   }
 
   # --------------------------------------------------------------------------#
-  # Configure Neighbor-level Address Family IPv4 Unicast
+  # Configure Neighbor-level Address Family IPv4 Unicast (default vrf)
   # --------------------------------------------------------------------------#
-  cisco_bgp_neighbor_af { '55.77 blue 1.1.1.1 ipv4 unicast':
+  if $operatingsystem == 'ios_xr' {
```
glennmatthews: I'd be tempted to remove the 'if' logic here and have these be set for both Nexus and XR. Nexus needs these configs too to have a 'working' config - it's just not as persnickety about validating the config as XR.

PR #189, opened 2016-02-11
Add secondary, broader search to facter platform lookup

chrisvanheuveln:
* NXOSv and possibly other platforms return an empty string to `facter -p cisco.hardware.type`, but they do respond to `facter -p cisco` with all cisco facts.
* This fix just adds a secondary lookup if the first attempt fails

### DIFF
--- a/tests/beaker_tests/lib/utilitylib.rb
+++ b/tests/beaker_tests/lib/utilitylib.rb
@@ -779,6 +779,9 @@ def platform
     # Some platforms do not respond correctly to the first command;
     # make another attempt using a broader search.
     on(agent, facter_cmd('-p cisco | egrep -A1 hardware'))
+    # Sample output:
+    #   hardware => {
+    #     type => "NX-OSv Chassis",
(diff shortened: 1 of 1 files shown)


tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -774,9 +774,16 @@ def platform
   return @cisco_hardware unless @cisco_hardware.nil?
   pi = on(agent, facter_cmd('-p cisco.hardware.type')).stdout.chomp
   if pi.empty?
-    fail 'Unable to query Cisco hardware type using the ' \
+    logger.debug 'Unable to query Cisco hardware type using the ' \
       "'cisco.hardware.type' custom factor key"
+    # Some platforms do not respond correctly to the first command;
+    # make another attempt using a broader search.
+    on(agent, facter_cmd('-p cisco | egrep -A1 hardware'))
```
mikewiebe: Can you add sample output?

chrisvanheuveln: Done!

PR #201, opened 2016-02-15
Bump metadata for 1.2.2 release

hunner:
The metadata was never updated for 1.2.1 but a tag was already made, so this PR bumps the metadata to 1.2.2 so that the package will build cleanly with the correct version.


hunner:
Mistakenly made against the develop branch instead of master.

PR #204, opened 2016-02-18
Updated to support Beaker 2.37.0

LuvCurves:


LuvCurves:
@chrisvanheuveln Here is the new PR.

tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -50,11 +50,10 @@
...
-def get_namespace_cmd(host, cmdstr, options)
+def get_namespace_cmd(host, cmdstr, _options)
   case host['platform']
   when /cisco/
-    agentvrf = options[:HOSTS][host.to_s.to_sym]['vrf']
-    return "sudo ip netns exec #{agentvrf} " + cmdstr
+    return cmdstr
   else
     return cmdstr
   end
```
chrisvanheuveln: Why keep the case around if you always `return cmdstr` regardless?

chrisvanheuveln: Actually, bigger problems. We noticed a couple of things in the beaker code: beaker appears to be keying off `cisco-5` for nexus platforms and `cisco-7` for XR platforms:
   https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/cisco.rb#L24
   https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/cisco.rb#L75
   https://github.com/puppetlabs/beaker/blob/master/spec/beaker/host/cisco_spec.rb#L49

...which is a problem since those numbers are somewhat irrelevant to XR and somewhat interchangeable on the nexus side. After discussing with Glenn we think it would be better to use a more explicit platform pattern to differentiate these two platform families; ie: `cisco-nexus-` and `cisco-ios_xr-`.
 
I think we need to get this resolved before we can merge your PR.

LuvCurves:
@chrisvanheuveln I have submitted a PR against beaker that changes the platform names from cisco-5 & cisco-7 to cisco-nexus & cisco-ios_xr.

You can see these changes here: 
https://github.com/LuvCurves/beaker/blob/BKR-705/lib/beaker/host/cisco.rb#L24
https://github.com/LuvCurves/beaker/blob/BKR-705/lib/beaker/host/cisco.rb#L75
https://github.com/LuvCurves/beaker/blob/BKR-705/spec/beaker/host/cisco_spec.rb#L49

Beaker is usually released every other Wednesday, so the next scheduled release is 2/24/16

glennmatthews:
See puppetlabs/beaker#1078

glennmatthews:
puppetlabs/beaker#1078 is a non-backwards-compatible change (`'cisco-7-x86_64'` is now rejected as an invalid platform.) Now that beaker 2.37 has been released with this change, we probably need to mark that as the new minimum version. I'm working on some changes along those lines and will open a PR.

glennmatthews:
Also, we may need to wait for beaker 2.38 - see my PR puppetlabs/beaker#1081 for some additional fixes that seem to be needed.

LuvCurves:
@glennmatthews Beaker 2.38.0 is out.

glennmatthews:
Thanks @LuvCurves. I'm going to close this PR as #233 supersedes it. I'll verify #233 with the 2.38 release and update there as needed.

PR #206, opened 2016-02-19
Doc changes for route-map/route-policy differences.

brturn:
My suggested doc changes.  Instead of polluting all the places where route-map is mentioned, I've moved the route-map/route-policy distinction to an OS-differences section, and linked the first use of "route-map" in each other section to this new section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -605,7 +605,7 @@ Install a backup path into the forwarding table and provide prefix 'independent 
 Enables the receive capability of additional paths for all of the neighbors under this address family for which the capability has not been disabled.  Valid values are true, false, or 'default'
 
 ##### `additional_paths_selection`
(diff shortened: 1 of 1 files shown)


chburket:
:+1:

README.md
```diff
@@ -2063,6 +2063,14 @@ Minimum Requirements:
   * Cisco Nexus 31xx, OS Version 7.0(3)I2(1), Environments: Bash-shell, Guestshell
   * Cisco Nexus 30xx, OS Version 7.0(3)I2(1), Environments: Bash-shell, Guestshell
 
+## Cisco IOS Differences
```
glennmatthews: Nitpick - change this from 'IOS' to 'OS' (and update the hyperlinks accordingly) - NX-OS is an operating system, but it's not IOS.

chrisvanheuveln:
:+1:

PR #207, opened 2016-02-20
Use correct provider

tuxmea:
The install manifest is working with Puppet 4 only (type system usage).
Puppet 4 AIO packages (like the one which will be installed on Cisco
devices) must use the 'puppet_gem' provider instead of the system 'gem'
provider.


glennmatthews:
:+1:

mikewiebe:
:+1:

PR #210, opened 2016-02-23
Add missing ACE documentation.

chrisvanheuveln:
* I also cleaned up the `desc` msgs in the type definition
* I noticed that the test_ace.rb redirect property is not supported on N7k. After the grpc merge to develop we can look at converting this to the new test format and easily skip this property.


chrisvanheuveln:
I'm withdrawing this one for now. I need to create a caveats table to handle the platform differences.

PR #211, opened 2016-02-23
Merge feature/grpc into develop

chburket:
There were only 3 merge conflicts, which I believe I have resolved properly.  I've run beaker tests against XR for bgp (all passed or skipped) and interface.  Still running down a problem with interface, but I thought I'd get this PR out there so people can review the changes.  I'll also be running the above beaker tests against NXOSv and will post the results when done.


docs/README-develop-beaker-scripts.md
```diff
@@ -135,6 +199,18 @@ HOSTS:
           user: devops
           password: devopspassword
 
+    xr-agent:
+        roles:
+            - agent
+        platform: cisco-7-x86_64
```
glennmatthews: The example hosts.cfg will need to be updated if/when puppetlabs/beaker#1066 is merged - just an FYI, not related directly to this PR.

glennmatthews:
:+1: :shipit:

chburket:
BGP beaker tests pass/skip properly for XR and NX-OS.  The interface test fails for both (in different ways).  Probably shouldn't hold up the merge for that, though.  We can fix it after.

mikewiebe:
:+1:

PR #213, opened 2016-02-24
Hotfix 1.2.3 - Documentation Only

mikewiebe:
### Added
- Download link for Nexus 5000 and Nexus 6000 Open Agent Container (OAC).
- OAC programmability guide links.
- Complete cisco_ace documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ The following table indicates which providers are supported on each platform. As
 | [cisco_aaa_authorization_login_exec_svc](#type-cisco_aaa_authorization_login_exec_svc) | ✅ | ✅ | ✅ | ❌ | ❌ | ❌ |
 | [cisco_aaa_group_tacacs](#type-cisco_aaa_group_tacacs) | ✅ | ✅ | ✅ | ❌ | ❌ | ❌ |
 | [cisco_acl](#type-cisco_acl) | ✅ | ✅ | ✅ | ❌ | ❌ | ❌ |
-| [cisco_ace](#type-cisco_ace) | ✅ | ✅ | ✅ | ❌ | ❌ | ❌ |
+| [cisco_ace](#type-cisco_ace) | ✅ | ✅ | ✅ | ❌* | ❌* | ❌* | * [caveats](#cisco_ace-caveats) |
 | [cisco_command_config](#type-cisco_command_config) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
 | [cisco_bgp](#type-cisco_bgp) | ✅ | ✅ | ✅ | ✅* | ✅* | ✅* | * [caveats](#cisco_bgp-caveats) |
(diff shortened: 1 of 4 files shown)


chrisvanheuveln:
:+1:

README.md
```diff
@@ -576,20 +576,259 @@ Permit or deny Fragments for ACL. Valid values are 'permit-all' and 'deny-all'
...
+| N56xx    | unsupported        | unsupported            |
+| N6k      | unsupported        | unsupported            |
+| N7k      | unsupported        | unsupported            |
+| IOS XR   | unsupported        | unsupported            |
+
+#### <a name="cisco_ace-caveats">Caveats</a>
+
+| Property | Caveat Description |
+|:--------|:-------------|
+| `http_method` | Not supported on N56xx, N6k, N7k |
```
glennmatthews: We already say above that the whole provider is unsupported on these platforms - why call out individual properties as well?

README.md
```diff
@@ -576,20 +576,259 @@ Permit or deny Fragments for ACL. Valid values are 'permit-all' and 'deny-all'
...
+Determines whether the config should be present or not on the device. Valid values are 'present' and 'absent'.
+
+#### Properties
+
+##### `action`
+The action to perform with this ACE. Valid values are keywords `permit` or `deny`.
+
+| Example
+|:--
+| `action => 'permit'`
```
glennmatthews: These examples take an already long provider and make it even longer - why are they needed/useful here when we don't have them for any of the other providers? I can see where they're useful for specific complex properties (dst_addr, packet_length) but for ones where the property takes just a keyword or a boolean, they seem redundant.

README.md
chrisvanheuveln: These platforms are coming in a few weeks. Is it hurting anything to have this information here early?

README.md
chrisvanheuveln: True, this format isn't there for other providers but one of the comments we got back from the TMEs was that they had difficulty figuring out how to use some of these properties; ie. we have examples in the type definitions but you have to look up each type file or do puppet describe to see that; so this was an attempt to get this documentation moving in a direction that included example usage.

Do we need examples for every property? Maybe not. I added them for completeness. I'd prefer to keep them but we can remove them if you feel strongly about it.

glennmatthews: Just trying to figure how we can keep this document from growing without bounds. 

    For simple properties like this one, I just don't see how:

Valid values are keywords `permit` or `deny`.

    is really more difficult to figure out than:

Valid values are keywords `permit` or `deny`.

| Example
|:--
| `action => 'permit'`

README.md
glennmatthews: Just seemed confusing to me without that context.

README.md
mikewiebe: @glennmatthews @chrisvanheuveln Can I go ahead with this merge or is there something that is a must change/fix here?

chrisvanheuveln: :+1:

glennmatthews: :+1: with gripes.

PR #214, opened 2016-02-24
Convert test_ace to new beaker format

chrisvanheuveln:
Convert test_ace to new beaker format
* Tested on N9k, N7k, N5k
 * N5k provider needs some work to handle 0001:0001::0001/128 ipv6 addr format; Mike already has some code to handle this so I've left a TBD in the test to follow up

### DIFF
--- a/tests/beaker_tests/cisco_acl/test_ace.rb
+++ b/tests/beaker_tests/cisco_acl/test_ace.rb
@@ -13,307 +13,136 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ###############################################################################
-# TestCase Name:
-# -------------
-# test_ace.rb
-#
-# TestCase Prerequisites:
-# -----------------------
-# This is a Puppet ACE resource testcase for Puppet Agent on Nexus devices.
-# The test case assumes the following prerequisites are already satisfied:
-#   - Host configuration file contains agent and master information.
-#   - SSH is enabled on the N9K Agent.
-#   - Puppet master/server is started.
-#   - Puppet agent certificate has been signed on the Puppet master/server.
-#
-# TestCase:
-# ---------
-# This ACE resource test verifies cisco_ace title patterns.
-#
-# The following exit_codes are validated for Puppet, Vegas shell and
-# Bash shell commands.
-#
-# Vegas and Bash Shell Commands:
-# 0   - successful command execution
-# > 0 - failed command execution.
-#
-# Puppet Commands:
-# 0 - no changes have occurred
-# 1 - errors have occurred,
-# 2 - changes have occurred
-# 4 - failures have occurred and
-# 6 - changes and failures have occurred.
-#
-# NOTE: 0 is the default exit_code checked in Beaker::DSL::Helpers::on() method.
-#
-# The test cases use RegExp pattern matching on stdout or output IO
-# instance attributes to verify resource properties.
-#
-###############################################################################
-
-# Require UtilityLib.rb paths.
 require File.expand_path('../../lib/utilitylib.rb', __FILE__)
 
-# -----------------------------
-# Common settings and variables
-# -----------------------------
-testheader = 'Resource cisco_ace'
-
-# 'tests' hash
-# Top-level keys set by caller:
-# tests[:master] - the master object
-# tests[:agent] - the agent object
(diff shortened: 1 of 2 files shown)


tests/beaker_tests/cisco_acl/test_ace.rb
```diff
@@ -13,307 +13,136 @@
...
-    'packet_length' => 'range 80 1000',
-    'time_range'    => 'my_range',
-    'log'           => 'true',
+tests[:seq_20_v6] = tests[:seq_20_v4].clone
+tests[:seq_20_v6][:title_pattern] = 'ipv6 beaker6 20'
+
+tests[:seq_30_v4] = {
+  desc:           'IPv4 Seq 30',
+  title_pattern:  'ipv4 beaker 30',
+  platform:       'n(3|9)k',
```
chburket: Do you still need the platform key here, if you are using the unsupported_properties method to also filter the properties?

chrisvanheuveln: Yes, but it's really just for efficiency. In this testcase the only properties I care about testing are `http_method` and `tcp_option_length` but I still have to specify `action, proto, src_addr, dst_addr` for these to be used. So if this is an N7k then unsupported_properties will strip out the properties I care about and all I'm left with are the basic commands I've already tested with in the previous testcases. By adding the :platform key I can skip the entire testcase.

chburket:
:+1:

tests/beaker_tests/cisco_acl/test_ace.rb
```diff
@@ -13,307 +13,136 @@
...
-    src_addr        => 'any',
-    dst_addr        => 'any',
-  ",
-  resource:       {
-    'action'   => 'deny',
-    'proto'    => 'tcp',
-    'src_addr' => 'any',
-    'dst_addr' => 'any',
+# Test hash test cases
+tests[:seq_5] = {
```
mikewiebe: Should this be `:seq_5_v4` to follow the same convention as the other tests?

tests/beaker_tests/cisco_acl/test_ace.rb
```diff
@@ -13,307 +13,136 @@
...
+    proto:         'tcp',
+    src_addr:      'any',
+    src_port:      'eq 40',
+    dst_addr:      'any',
+    dst_port:      'range 32 56',
+
+    established:   'true',
+    packet_length: 'range 80 1000',
+    precedence:    'flash',
+    redirect:      'Ethernet1/1,Ethernet1/2,port-channel1',
```
mikewiebe: Hardcoded eth interfaces... eww

tests/beaker_tests/cisco_acl/test_ace.rb
chrisvanheuveln: I just didn't see the need. I can change it.

tests/beaker_tests/cisco_acl/test_ace.rb
```diff
@@ -13,307 +13,136 @@
...
-  ",
-  resource:       {
-    'action'   => 'deny',
-    'proto'    => 'tcp',
-    'src_addr' => 'any',
-    'dst_addr' => 'any',
+tests[:seq_10_v6] = {
+  desc:           'IPv6 Seq 10',
+  title_pattern:  'ipv6 beaker6 10',
+  platform:       'n(3|7|9)k', # TBD: N5k addrs display as 0001:0001::0001/128
```
mikewiebe: I sent you the link for this.. should we go ahead and take care of it now or do you want to defer?

tests/beaker_tests/cisco_acl/test_ace.rb
chrisvanheuveln: Yes. Oddly enough it will accept `Ethernet999/99` but it will not nvgen it if it doesn't exist; however, it will nvgen `port-channel` regardless. So I'll just change it to specify port-channels and it should be fine.

tests/beaker_tests/cisco_acl/test_ace.rb
chrisvanheuveln: defer. I expect this one may not be as straightforward.

PR #216, opened 2016-02-25
Provider bulk rename nxapi -> cisco

glennmatthews:
Complete the work that started in feature/grpc - since our providers are no longer NXAPI-exclusive, rename all of the providers from 'nxapi' to 'cisco'.

* Move all `lib/puppet/provider/*/nxapi.rb` --> `lib/puppet/provider/*/cisco.rb`
* Replace all `:nxapi` symbols in these files to `:cisco` (provider symbol must match filename)
* Update package test, docs, and demo manifest to use  `provider => 'cisco'` instead of `provider => 'nxapi'`
* Remove `cisco_nxapi` fact as we don't depend on that gem any more (it's now part of cisco_node_utils)
* Remove checksums.json file as it is stale and serves no purpose in the repo (it's dynamically generated by 'puppet module build' as needed)
* Update provider templates and docs to reflect these changes.
* Update/remove all other references to `nxapi` or `NXAPI` (mostly header comments and provider `desc` strings).


chrisvanheuveln:
:+1:

checksums.json
```diff
@@ -1,147 +0,0 @@
-{
```
robert-w-gries: What is this file?  Should we exclude it in our gitignore?

glennmatthews: When you build a puppet module the resulting .tar.gz file includes a generated checksums.json that contains the checksums of all files in the module. The file doesn't need to be in the repository as it's dynamically generated. This file hadn't been touched since the very first release - I think it got added by mistake at that point - so I don't think we need to add it to .gitignore unless it actually shows up again.

robert-w-gries: If the file is generated every time we build the module then I would prefer it be added to the gitignore.  I use `git add --all` when staging files, and it'd be easy for `checksums.json` to slip through the cracks if I was committing a large number of changes across many files.

glennmatthews: It's not added to this directory though - it's only created inside the `puppetlabs-ciscopuppet-1.2.3.tar.gz` file. I just checked this by rerunning `puppet module build` and the file is not regenerated in this repo. We already exclude the build directory (`/pkg`), so the only way this would creep back in would be for you to unpack the module into this directory.

robert-w-gries: Ah I see, alright then the PR looks good to merge :+1:

PR #223, opened 2016-03-07
Cisco package provider for XR

sushirol:


lib/puppet/provider/package/cisco.rb
```diff
@@ -17,10 +17,10 @@
 require 'puppet/util/package'
 
 Puppet::Type.type(:package).provide :cisco, parent: :yum do
-  desc "The cisco nexus package provider.
+  desc "The cisco package provider.
   Local rpm installations will utilize the native yum provider.
   Cisco rpm installations from the host will utilize the native yum provider.
-  Cisco rpm installations from guestshell will utilize nxapi to install to host."
+  Cisco rpm installations from guestshell will utilize api to install to host."
```
glennmatthews: I don't think this is correct. Probably this should be something like:

```diff
-  desc "The cisco nexus package provider.
+  desc "The cisco package provider.
   Local rpm installations will utilize the native yum provider.
-  Cisco rpm installations from the host will utilize the native yum provider.
+  Cisco NX-OS rpm installations from the host will utilize the native yum provider.
-  Cisco rpm installations from guestshell will utilize nxapi to install to host."
+  Cisco NX-OS rpm installations from guestshell will use this provider to install to host via NXAPI.
+  Cisco IOS XR rpm installations will utilize this provider to install to host via sdr_instcmd."
```

lib/puppet/provider/package/cisco.rb
```diff
@@ -160,7 +166,7 @@ def in_guestshell?
 
   def install
     if in_guestshell? && target_host?
-      debug 'Guestshell + target=>host detected, using nxapi for install'
+      debug 'Guestshell + target=>host detected, using api for install'
```
glennmatthews: This should stay NXAPI.

lib/puppet/provider/package/cisco.rb
```diff
@@ -160,7 +166,7 @@ def in_guestshell?
 
   def install
```
glennmatthews: I don't see where this method is calling XR-specific logic - it seems like it would fall through to the 'else' case below and try and use 'yum' to install the package which is wrong, right?

lib/puppet/provider/package/cisco.rb
```diff
@@ -178,11 +184,11 @@ def install
 
   def uninstall
```
glennmatthews: Same comment here - looks like XR uninstall would be using yum.uninstall instead of Cisco::Yum.uninstall.

sushirol:
@glennmatthews @chburket I have pushed a new set of changes. Ready for a review.

lib/puppet/provider/package/cisco.rb
```diff
@@ -159,8 +208,13 @@ def in_guestshell?
...
+    if (in_nexus? && in_guestshell? && target_host?) ||
+       (in_ios_xr? && target_host?)
+      if in_nexus?
+        debug 'Guestshell + target=>host detected, using nxapi for install'
+      elsif in_ios_xr?
+        debug 'target=>host detected, using sdr_instcmd for install'
+      end
       if @resource[:source]
         Cisco::Yum.install(@resource[:source])
       else
```
glennmatthews: Have you tested both conditional branches (with and without `@resource[:source]`) on IOS XR?

lib/puppet/provider/package/cisco.rb
```diff
@@ -53,7 +66,8 @@
   # this method in package.rb determines how to set ensure
   # @property_hash is empty at this point
   def properties
-    if in_guestshell? && target_host?
+    if (in_nexus? && in_guestshell? && target_host?) ||
+       (in_ios_xr? && target_host?)
```
glennmatthews: target_host? is used to indicate that guestshell should install to the host (native NX) instead of within the guestshell LXC. It doesn't really apply to XR as a concept at present. So this could probably just be:

```ruby
if in_ios_xr? || (in_guestshell && target_host?)
```

lib/puppet/provider/package/cisco.rb
```diff
@@ -159,8 +208,13 @@ def in_guestshell?
   end
 
   def install
-    if in_guestshell? && target_host?
-      debug 'Guestshell + target=>host detected, using nxapi for install'
+    if (in_nexus? && in_guestshell? && target_host?) ||
+       (in_ios_xr? && target_host?)
```
glennmatthews: Ditto above comment about target_host? not being applicable to XR.

glennmatthews:
You will probably want to update docs/README-package-provider.md and CHANGELOG.md.

glennmatthews:
Also, do you plan to update the tests/beaker_tests/file_service_package/package_provider_nondefaults_2.rb test script to add XR support?

lib/puppet/provider/package/cisco.rb
```diff
@@ -30,7 +31,7 @@
   # these commands must exist to execute native yum provider
   commands yum: 'yum', rpm: 'rpm', python: 'python'
 
-  defaultfor operatingsystem: :nexus
+  defaultfor operatingsystem: [:ios_xr, :nexus]
```
HelenHZhang: Reading puppet type and provider book, it seems we do not need to set "defaultfor" at all.

PR #231, opened 2016-03-16
Documentation updates for IOS XR

glennmatthews:
* Add setup/installation examples for IOS XR
* Move information about installing/configuring cisco_node_utils gem into a new README-gem-install.md document so that this information can be easily referenced from several of the other READMEs without duplication.
* Various other attempts at reducing the duplication of information amongst multiple documents.

### DIFF
--- a/docs/README-agent-install.md
+++ b/docs/README-agent-install.md
@@ -1,4 +1,4 @@
-# Puppet Agent Installation & Setup: Cisco Nexus
+# Puppet Agent Installation & Setup: Cisco Nexus and Cisco IOS XR
 
 #### Table of Contents
 
@@ -15,7 +15,7 @@
 
 ## <a name="overview">Overview</a>
 
-This document describes Puppet agent installation and setup on Cisco Nexus switches. These instructions focus on manual setup. See the [Automated Installation](#auto-install) section for documentation regarding alternative installation methods.
+This document describes Puppet agent installation and setup on Cisco Nexus switches and Cisco IOS XR routers. These instructions focus on manual setup. See the [Automated Installation](#auto-install) section for documentation regarding alternative installation methods.
 
 ![1](puppet_outline.png)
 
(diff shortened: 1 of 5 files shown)


docs/README-gem-install.md
```diff
@@ -0,0 +1,79 @@
...
+This gem has various dependencies which differ between IOS XR and Nexus; installing `cisco_node_utils` by itself will automatically install the dependencies that are relevant to the target platform.
+
+Nexus example:
+
+~~~bash
+[root@guestshell]# /opt/puppetlabs/puppet/bin/gem install cisco_node_utils
+...
+[root@guestshell]# /opt/puppetlabs/puppet/bin/gem list | egrep 'cisco|net_http'
+cisco_node_utils (1.2.0)
+net_http_unix (0.2.1)
```
robert-w-gries: Should probably remove this reference since we're deprecating `net_http_unix`

docs/README-agent-install.md
```diff
@@ -24,18 +24,21 @@ This document describes Puppet agent installation and setup on Cisco Nexus switc
 #### Platform and Software Minimum Requirements
 
 * The Cisco NX-OS Puppet implementation requires open source Puppet version 4.0 or Puppet Enterprise 2015.2
+* The Cisco IOS XR Puppet implementation requires open source Puppet version 4.3.2 or Puppet Enterprise 2015.3.2
```
robert-w-gries: Note: our `metadata.json` does not reflect the same Puppet Enterprise version requirement.

```
  "requirements": [
    {
      "name": "pe",
      "version_requirement": "2015.2.x"
    },
```

glennmatthews: Doesn't that match up with the minimum version for NX-OS on the line above?

docs/README-gem-install.md
glennmatthews: Eh? AFAIK the NXAPI client still requires net_http_unix.

robert-w-gries: Woops thinking of the wrong gem :-)

docs/README-agent-install.md
robert-w-gries: It does.. but our `metadata.json` doesn't handle any platform differences in version requirements and in fact doesn't mention XR at all.  If XR has the minimum requirements of Puppet 4.3.2 and PE 2015.3.2, then our `metadata.json` should reflect that.  This isn't a necessary change for your PR, but it's something I wanted to clarify.

robert-w-gries:
:+1:

PR #234, opened 2016-03-18
modify switchport_trunk_allowed_vlan to use range_summarize()

dcheriancisco:
This change modifies the type file of cisco_interface for the the property switchport_trunk_allowed_vlan to accept an arbitary range strings with the use of range_summarize(). The utility method range_summarize() massages the input resource string to a form that is compatible with the range summarization that is done by NxOS CLI on VLAN ranges. This takes care of idempotency issues.


chrisvanheuveln:
Changes look good. Are there any changes needed in NU? 

Also, please update the switchport_trunk_vlan_allowed entry in README.md to indicate the new syntax support.

chburket:
:+1:

PR #244, opened 2016-04-05
bgp: beaker platform fixes #1

chrisvanheuveln:
* bgp: reconnect_interval was missing from test
* bgp_af: change redist expected value to sorted
* bgp_neighbor: add log_neighbor_changes

### DIFF
--- a/tests/beaker_tests/cisco_bgp_af/test_bgpaf.rb
+++ b/tests/beaker_tests/cisco_bgp_af/test_bgpaf.rb
@@ -208,7 +208,7 @@ tests[:non_def_N] = {
   },
 }
 
-redistribute = Array[%w(static RouteMap), %w(lisp RouteMap)]
+redistribute = Array[%w(lisp RouteMap), %w(static RouteMap)]
 tests[:non_def_R] = {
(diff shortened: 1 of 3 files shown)


tests/beaker_tests/cisco_bgp_af/test_bgpaf.rb
```diff
@@ -208,7 +208,7 @@
   },
 }
 
-redistribute = Array[%w(static RouteMap), %w(lisp RouteMap)]
+redistribute = Array[%w(lisp RouteMap), %w(static RouteMap)]
```
robert-w-gries: Why is the swap necessary?

chrisvanheuveln: The redist getter return values were different on n6k (lex.sort) vs n9k (unsorted) so I added a sort to the NU getter to normalize the data.
The array above is used by puppet resource so it has to be in sorted order to match the getter output.

robert-w-gries:
:+1:

chburket:
:+1:

PR #250, opened 2016-04-11
fabric_path changes

saichint:
This is for fabricpath, interface_channel_group and interface_portchannel
All the related beaker tests and demo_manifests passed.
The README.md file is changed to include the matrix for provider compliance.
In the beaker_test file defaults for n7k 2 params have to be removed as these are removed from node_utils before (fabricpath)
For interface_portchannel the default for lacp_max_bundle should be 32 (same as n9k)


README.md
```diff
@@ -139,10 +139,12 @@ The following table indicates which providers are supported on each platform. As
 | [cisco_bridge_domain_vni](#type-cisco_bridge_domain_vni) | ❌ | ❌ | ❌ | ❌ | ❌ | ✅ | ❌ | ❌ |
 | [cisco_encapsulation](#type-cisco_encapsulation) | ❌ | ❌ | ❌ | ❌ | ❌ | ✅ | ❌ | ❌ |
 | [cisco_evpn_vni](#type-cisco_evpn_vni) | ✅ | ❌ | ❌ | ✅ | ✅ | ✅ | ✅ | ❌ |
+| [cisco_fabricpath_global](#type-cisco_fabricpath_global) | ❌ | ❌ | ❌ | ✅ | ✅ | ✅ | ❌ | ❌ | * [caveats](#cisco_fabricpath_global-caveats) |
```
mikewiebe: Remind me again why we don't support fabricpath on n3k, n9k and n8k?

mikewiebe:
One question, but other then that :+1:

README.md
saichint: Deepak told me that fabricpath uses own our asics for some patented functionality and n3k,n9k and n8k use broadcom chips which don't have this.

PR #252, opened 2016-04-12
cisco_vtp fixes.

mikewiebe:
**Summary of changes:**
- Updated docs to indicate support on n5k, n6k and n7k.
- Append `bootflash:/` to vtp filename if now mount is specified.

Tested on: n3k, n7k, n8k, n9k(I2 & I3).  Waiting on n5|6k to become available.

Note, all tests pass except an idempotence issue for vtp password.  This is a platform issue and the work around will be in a separate PR.

### DIFF
--- a/lib/puppet/type/cisco_vtp.rb
+++ b/lib/puppet/type/cisco_vtp.rb
@@ -74,10 +74,17 @@ Puppet::Type.newtype(:cisco_vtp) do
   newproperty(:filename) do
     desc "VTP file name. Valid values are string, keyword 'default'."
 
-    munge do |file_name|
-      file_name = :default if file_name == 'default'
+    validate do |file_name|
       fail 'File name is not a string.' unless
         file_name == :default || file_name.is_a?(String)
+    end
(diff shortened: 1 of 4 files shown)


lib/puppet/type/cisco_vtp.rb
```diff
@@ -74,10 +74,17 @@
...
-    munge do |file_name|
-      file_name = :default if file_name == 'default'
+    validate do |file_name|
       fail 'File name is not a string.' unless
         file_name == :default || file_name.is_a?(String)
+    end
+
+    munge do |file_name|
+      file_name = :default if file_name == 'default'
+      unless file_name == :default || /(bootflash:|usb.*:)/.match(file_name)
```
chrisvanheuveln: Nit: _maybe_ a little cleaner to do:

```
# Prefix bootflash:/ if only given a filename
file_name.insert(0, 'bootflash:/') unless file_name[/\//]

```

...but what you have is fine too. :+1: either way

PR #253, opened 2016-04-12
Fix things for itd, stp, snmp_comm

saichint:
These are changes for 3 providers.
1, fcoe attribute in stp_global is not supported on n3k and so it is removed from demo manifest and beaker tests.
2. itd is supported only on n9k and n7k. Added exclude clause in the demo manifest so that the manifest will not throw errors for other nexus switches.
3. snmp_commity provider has an error where if there is a comminuty already existing on the switch, the beaker test is failing when it is trying to remove (because it is accessing group parameter after the community resource is removed). This is fixed.

### DIFF
--- a/examples/cisco/demo_itd.pp
+++ b/examples/cisco/demo_itd.pp
@@ -18,183 +18,185 @@
 
 class ciscopuppet::cisco::demo_itd {
 
-  $nat_destination = platform_get() ? {
-    'n7k'   => true,
-    default => undef
-  }
+  if platform_get() =~ /n(7|9)k/ {
+    $nat_destination = platform_get() ? {
+      'n7k'   => true,
+      default => undef
+    }
 
-  $peer_local1 = platform_get() ? {
-    'n9k'   => 'pser1',
-    default => undef
-  }
+    $peer_local1 = platform_get() ? {
+      'n9k'   => 'pser1',
+      default => undef
+    }
 
-  $peer_local2 = platform_get() ? {
-    'n9k'   => 'pser2',
-    default => undef
-  }
+    $peer_local2 = platform_get() ? {
+      'n9k'   => 'pser2',
+      default => undef
+    }
 
-  $peer_vdc1 = platform_get() ? {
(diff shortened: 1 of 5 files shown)


examples/cisco/demo_itd.pp
```diff
@@ -18,183 +18,185 @@
...
+    cisco_acl { 'ipv4 ial':
+      ensure => present
+    }
+  
+    cisco_acl { 'ipv4 eal':
+      ensure => present
+    }
+
+    cisco_interface { 'ethernet1/1':
+      switchport_mode => disabled
```
chrisvanheuveln: Can you reference `disabled` without quotes in the manifest? If so I was unaware of this and I'm wondering if we should add the quotes as a best practice.

examples/cisco/demo_itd.pp
```diff
@@ -18,183 +18,185 @@
...
+      load_bal_enable               => true,
+      load_bal_buckets              => 16,
+      load_bal_mask_pos             => 10,
+      load_bal_method_bundle_hash   => 'ip',
+      load_bal_method_bundle_select => 'dst',
+      peer_vdc                      => $peer_vdc2,
+      peer_local                    => $peer_local2,
+      shutdown                      => false,
+    }
   }
```
chrisvanheuveln: Please add an else condition at line 201:

```
-  }
+  } else {
+    notify{'SKIP: This platform does not support cisco_itd_service': }
+  }
```

chrisvanheuveln:
:+1: Just a couple of minor comments. Please address and I'll merge after that.

examples/cisco/demo_itd.pp
saichint: I do see that in interface.pp at:
https://github.com/cisco/cisco-network-puppet-module/blob/develop/examples/cisco/demo_interface.pp#L41

Anyway, I changed it the way you asked but it it was like this always I believe

chrisvanheuveln: yup, not saying it's wrong, it just caught my eye as unusual since disabled is not a keyword.

PR #255, opened 2016-04-13
Feature/aaa fixup release

robert-w-gries:
* Remove vshell commands from tests
* Remove platform check that failed for non-3k and non-9k platforms
* Fix `authentication_login` test that rotated values the incorrect way

Tests passing all platforms:

* `cisco_aaa_group_tacacs`
* `cisco_aaa_authentication_login`
* `cisco_aaa_login_cfg_svc`
* `cisco_aaa_login_exec`

### DIFF
--- a/tests/beaker_tests/cisco_aaa_authentication_login/test_aaaauthelogin_nondefaults.rb
+++ b/tests/beaker_tests/cisco_aaa_authentication_login/test_aaaauthelogin_nondefaults.rb
@@ -94,7 +94,7 @@ test_name "TestCase :: #{testheader}" do
     tests[id][:resource_cmd] =
       get_namespace_cmd(agent, resource_cmd_str, options)
 
-    tests[id][:code] = [0, 2]
+    tests[id][:code] = [2]
     tests[id][:desc] =
       "1.1 Apply manifest with non-default #{prop}, and test harness"
     create_aaaauthelogin_manifest(tests, id)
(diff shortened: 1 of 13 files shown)


tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -268,7 +268,7 @@ def test_resource(tests, id, state=false)
 # Wrapper for config pattern-match tests
 def test_show_cmd(tests, id, state=false)
   stepinfo = format_stepinfo(tests, id, 'SHOW CMD')
-  show_cmd = get_vshell_cmd(tests[:show_cmd])
+  show_cmd = get_vshell_cmd(tests[id][:show_cmd])
```
chrisvanheuveln: :+1: on your changes as is; just wondering if we can take this opportunity to get rid of `test_show_cmd` altogether at this point? Looks like only one test might actually still be using this and it's a aaa test...

`grep  test_show_cmd tests/* -r`
`grep grep show_pattern tests/* -r`

chburket:
:+1:

tests/beaker_tests/lib/utilitylib.rb
robert-w-gries: Done

robert-w-gries:
Is it preferred to have the doc updates in this PR or in a separate one?

tests/beaker_tests/cisco_aaa_authentication_login/test_aaaauthelogin_nondefaults.rb
```diff
@@ -97,12 +96,12 @@
 
     tests[id][:code] = [0, 2]
```
mikewiebe: I know this is not part of your changes, but I would expect the exit code to be 2 each time for this test if the test is cleaning up properly in the setup for the test.

chrisvanheuveln:
doc updates: Your call. I like smaller reviews :-)

tests/beaker_tests/cisco_aaa_authentication_login/test_aaaauthelogin_nondefaults.rb
robert-w-gries: Good catch.  Test still passes with the change

mikewiebe:
:+1:

PR #257, opened 2016-04-14
Update supported aaa platforms in readme

robert-w-gries:


mikewiebe:
Please update the changelog to indicate new platform support for aaa.

Example: https://github.com/cisco/cisco-network-puppet-module/blob/release_1.3.0/CHANGELOG.md#added

robert-w-gries:
Updated changelog

PR #258, opened 2016-04-14
Rel/fp

saichint:
This is for vpc, name_server of netdev (and portchannel_global changes requested by Mike) 
1. README.md changed to remove vpc support for n8k (and add support for netdev)
2. demo_vpc_plus.pp has now only n7k support. Deepak said there is no support for n5k or n6k for vpc plus. (The demo_manitests are failing for attrubutes which are excluded in the NU but not in manifest fir n5/6k). The platform check is left in regex format and also another platform check for n7k vs others are left as they were, as requested by Deepak, because once the support for n5/6k is in, all we need to change a single line of manifest.
3. The order of set attributes is causing issues and 2 variables are getting overwritten and this is fixed now in the provider file. Basically, auto_recovery, peer_gateway need to be set first before any non-bool variables. The other bool variables have to be set after the non-bool variables.
4. For portchannel_global, "fail arg" cases are added to ensure that if wrong type of arguments are present or required arguments are absent in the manifest, the sets will not go through.

### DIFF
--- a/examples/cisco/demo_vpc_plus.pp
+++ b/examples/cisco/demo_vpc_plus.pp
@@ -16,7 +16,7 @@
 
 class ciscopuppet::cisco::demo_vpc_plus {
 
-  if platform_get() =~ /n(5|6|7)k/ {
+  if platform_get() =~ /n7k/ {
     if platform_get() == 'n7k' {
       $fabricpath_multicast_load_balance = true
       $port_channel_limit = false
--- a/examples/cisco/demo_vpc_plus.pp
+++ b/examples/cisco/demo_vpc_plus.pp
@@ -17,6 +17,7 @@
 class ciscopuppet::cisco::demo_vpc_plus {
 
   if platform_get() =~ /n7k/ {
+    # this next block is for future use when n5k and n6k support gets added
     if platform_get() == 'n7k' {
       $fabricpath_multicast_load_balance = true
       $port_channel_limit = false
--- a/lib/puppet/provider/cisco_vpc_domain/cisco.rb
+++ b/lib/puppet/provider/cisco_vpc_domain/cisco.rb
@@ -51,22 +51,26 @@ Puppet::Type.type(:cisco_vpc_domain).provide(:cisco) do
     :system_mac,
     :system_priority,
   ]
-  VPC_BOOL_PROPS = [
+  VPC_BOOL_PROPS1 = [
     :auto_recovery,
+    :peer_gateway,
+  ]
+  VPC_BOOL_PROPS2 = [
     :fabricpath_multicast_load_balance,
     :graceful_consistency_check,
     :layer3_peer_routing,
-    :peer_gateway,
     :port_channel_limit,
     :self_isolation,
     :shutdown,
   ]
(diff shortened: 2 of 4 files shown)


examples/cisco/demo_vpc_plus.pp
```diff
@@ -16,7 +16,7 @@
 
 class ciscopuppet::cisco::demo_vpc_plus {
 
-  if platform_get() =~ /n(5|6|7)k/ {
+  if platform_get() =~ /n7k/ {
     if platform_get() == 'n7k' {
       $fabricpath_multicast_load_balance = true
       $port_channel_limit = false
```
chrisvanheuveln: This if-else block would appear to be unnecessary now.

lib/puppet/provider/cisco_vpc_domain/cisco.rb
```diff
@@ -51,22 +51,26 @@
     :system_mac,
     :system_priority,
   ]
-  VPC_BOOL_PROPS = [
+  VPC_BOOL_PROPS1 = [
     :auto_recovery,
+    :peer_gateway,
+  ]
```
chrisvanheuveln: Please add a comment to explain why two bool arrays are needed to ensure the setter processing order.

chrisvanheuveln:
Please use a more descriptive title for your PRs. :-)

examples/cisco/demo_vpc_plus.pp
saichint: This one I left it because Deepak asked me, he says that once the support for 5k ans 6k is there, we just need to change this line rather than remove it altogether. we can discuss in today's meeting

chrisvanheuveln: You can keep it.

saichint:
Added comment to demo file. If you can give me the corrections for the provider file, I can change them properly

lib/puppet/provider/cisco_vpc_domain/cisco.rb
```diff
@@ -51,10 +51,15 @@
...
   ]
+  # auto_recovery and peer_gateway need to be set before non_boolean
+  # properties like auto_recovery_reload_delay and peer_gateway_exclude_vlan
+  # so making it into a new array
   VPC_BOOL_PROPS1 = [
     :auto_recovery,
     :peer_gateway,
   ]
+  # these properties like port_channel_limit need to be set after non_boolean
+  # properties like fabricpath_emulated_switch_id
```
chrisvanheuveln: I think a single comment is all you need here. How about this instead:

```
  # There are multiple BOOL arrays due to dependency requirements. The order of
  # these property arrays is important for proper processing by properties_set.
  VPC_BOOL_PROPS1 = [
    :auto_recovery,
    :peer_gateway,
  ]
   VPC_BOOL_PROPS2 = [
      :fabricpath_multicast_load_balance,
      :graceful_consistency_check,
     :layer3_peer_routing,
     :port_channel_limit,
      :self_isolation,
      :shutdown,
    ]
  VPC_PROPS = VPC_BOOL_PROPS1 + VPC_NON_BOOL_PROPS + VPC_BOOL_PROPS2
```

That's all I was looking for. :-)

PR #259, opened 2016-04-15
test_interface_service_vni: platform cleanups

chrisvanheuveln:
* Fixes for interface_service_vni and encapsulation

* Add 'default' support to encapsulation and fixed idempotence issues

* Add pre-reqs and skips to demo manifests and beaker tests

* Reformatted beaker tests to use latest syntax

* Tested on n7k, all other platforms are skipped

### DIFF
--- a/examples/cisco/demo_interface_service_vni.pp
+++ b/examples/cisco/demo_interface_service_vni.pp
@@ -15,12 +15,27 @@
 # limitations under the License.
 
 class ciscopuppet::cisco::demo_interface_service_vni {
-  cisco_interface_service_vni { 'Ethernet9/2 344' :
-    encapsulation_profile_vni   => 'vni_500_5000',
-    shutdown                    => true,
-  }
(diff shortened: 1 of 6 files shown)


examples/cisco/demo_interface_service_vni.pp
```diff
@@ -15,12 +15,27 @@
...
-    encapsulation_profile_vni   => 'vni_600_6000',
-    shutdown                    => 'default',
+
+  if platform_get() =~ /n7k/ {
+
+    cisco_encapsulation {"vni_500_5000" :
+      ensure          => present,
+      dot1q_map       => ['500', '5000'],
+    }
+    cisco_interface_service_vni { 'Ethernet9/2 344' :
```
mikewiebe: Is this the test that needs a special linecard?  If so, should we be checking for that?

chrisvanheuveln: I knew that function would come back to haunt me. :-) 
I'll add it.

PR #268, opened 2016-04-20
Feature/vxlan bridge domain

rkorlepa:
This is the beaker and demo manifest for interface bdi.

Have verified this manifest and beaker on n3k, n5k, n6k , n7k , n8k, n9k platforms.
This can be created only in n7k so it will skip on all other platforms.
This also contains platform check in test_bridge_domain and test_bridge_domain_vni which was missing before.

interface bdi100
  ip address 10.10.10.1/24
  vrf member Coke
  no shutdown

cisco_interface { 'Bdi100':
  require               => Cisco_bridge_domain['100'],
  ensure                => 'present',
  shutdown              => false,
  ipv4_address          => "10.10.10.1",
  ipv4_netmask_length   => 24, 
  vrf                   => 'Coke'
}


examples/cisco/demo_interface.pp
```diff
@@ -88,6 +88,21 @@
       svi_management   => true,
       ipv4_arp_timeout => 300,
     }
+    
+    cisco_bridge_domain { "100":
+      ensure    => 'present',
+      shutdown  => false,
+      bd_name   => 'Coke'
```
chrisvanheuveln: Please don't use real company names in our code or examples.

tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -1029,3 +1029,10 @@ def find_interface(tests, id=nil, skipcheck=true)
   end
   intf
 end
+
+def create_bridge_domain(tests)
+  # BDI_non_default test case requires to create a bridge domain
+  return unless platform == 'n7k'
+  bd = tests[:bdi_name][/(\d+)/]
+  config_bridge_domain(agent, bd)
+end
```
chrisvanheuveln: `utilitylib.rb` is for common methods so this shouldn't be here unless you intend to use it in other test files. 
But I don't think you need it at all. Just do this in `test_interface.rb` instead:

```
+  if platform_supports_test(tests, 'BDI_non_default')
     logger.info("\n#{'-' * 60}\nSection 6. BDI Property Testing")
     bd = tests[:bdi_name][/(\d+)/]
-    create_bridge_domain(tests)
+    config_bridge_domain(agent, bd)
     test_harness_interface(tests, 'BDI_non_default')
+  end
```

examples/cisco/demo_interface.pp
rkorlepa: Done changed

tests/beaker_tests/lib/utilitylib.rb
rkorlepa: Moved it to test_interface.rb file

PR #272, opened 2016-04-20
netdev/network_interface: refactor beaker test

chrisvanheuveln:
* Fixed problem in provider where it didn't handle dynamic interfaces
* Refactored the beaker test to the new template style
* Speed/Duplex/Mtu now probe the test interface using interface_capabilities; also pre-test with variables to see which commands actually work on a given linecard, since there is so much variability among platforms/linecards.
* Tested on n3k,n6k,n7k,n8k(skips no LCs),n9k


tests/beaker_tests/netdev_stdlib/test_network_interface.rb
```diff
@@ -0,0 +1,144 @@
...
+  precheck_testable_non_defaults(tests, resource_cmd, cap)
+  precheck_testable_auto(tests, resource_cmd)
+  interface_cleanup(agent, intf, 'Post-Pre-Check Cleanup')
+end
+
+# Discover testable non-default property values
+def precheck_testable_non_defaults(tests, cmd, cap)
+  agent = tests[:agent]
+  testable = {}
+  cap['Speed'].to_s.split(',').each do |cap_speed|
```
mikewiebe: Would it be useful to move the auto-discovery logic bits for speed and duplex moved into `utilitylib.rb` as more generic helper methods?  That way it can be used by other provider tests like `cisco_interface` etc...

chrisvanheuveln: Sure. I'll rework them as generic.

tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -1032,3 +1033,80 @@ def interface_capabilities(agent, intf)
...
+#     cmd:         '<PUPPET_BINPATH> resource network_interface ',
+#     intf:        'ethernet1/1',
+#     caps:        {'Speed' => '10,100,1000', 'Duplex' => 'auto,half,full'},
+#     probe_props: %w(Speed Duplex)
+#
+def interface_probe(tests, probe={})
+  agent = tests[:agent]
+
+  # Use tests[:probe] if caller does not supply a separate probe hash
+  probe = tests[:probe] if probe.empty?
```
robert-w-gries: Add check for nil `tests[:probe]`

robert-w-gries:
👍 Great addition

I believe the virtual Frettas have interfaces so can you test against one of them before we merge?

chrisvanheuveln:
I did test with N8k already. It works. I see that I messed up debug_probe after I put it in its own method so I need to fix that before merge.

PR #277, opened 2016-04-25
Release 1.3.0

dcheriancisco:
Adding new custom facts to cisco facter params and using it in beaker


lib/facter/cisco.rb
```diff
@@ -43,6 +44,10 @@
 
     hash['virtual_service'] = Platform.virtual_services
 
+    hash['feature_compatible_module_iflist'] = {}
+    interface_list = Feature.compatible_interfaces('fabricpath')
+    hash['feature_compatible_module_iflist']['fabricpath'] = interface_list
```
chrisvanheuveln: I like what you've done with this however I worry about adding this lookup tax to every single facter run. I had the same concern with interface_capabilities (which is only used for beaker) and ultimately decided it was better in that case to keep as an on-demand call.

In this case it looks fairly lightweight and could possibly be used by some of the providers themselves.

tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -792,10 +783,11 @@ def mt_full_interface
...
-  cmd = get_vshell_cmd('sh mod')
-  out = on(agent, cmd, pty: true).stdout[/^(\d+)\s.*N7K-F(?:2.*25E|3)/]
-  slot = out.nil? ? nil : Regexp.last_match[1]
-  "ethernet#{slot}/1" unless slot.nil?
+  cmd = '-p cisco.feature_compatible_module_iflist.fabricpath'
+  if_array_str = on(agent, facter_cmd(cmd)).stdout.chomp
+  if_array_str.gsub!(/[\[\]\n\s"]/, '')
+  if_array = if_array_str.split(',')
+  if_array[0] unless if_array.empty?
 end
```
chrisvanheuveln: For post-release, consider changing this method from `fabricpath_interface` to `compatible_interface(feature)` since we'll probably need this for other linecard-specific providers.

chrisvanheuveln:
👍

tests/beaker_tests/lib/utilitylib.rb
mikewiebe: +1

mikewiebe:
:+1:

PR #281, opened 2016-04-26
netdev network_vlan

saichint:
Added code to get rid of bridge domains so that the beaker tests pass

### DIFF
--- a/tests/beaker_tests/netdev_stdlib/network_vlan_provider_nondefaults.rb
+++ b/tests/beaker_tests/netdev_stdlib/network_vlan_provider_nondefaults.rb
@@ -64,6 +64,8 @@ test_name "TestCase :: #{testheader}" do
   step 'TestStep :: Setup switch for provider test' do
     resource_absent_cleanup(agent, 'network_vlan',
                             'Setup switch for network_vlan provider test')
+    resource_absent_cleanup(agent, 'cisco_bridge_domain',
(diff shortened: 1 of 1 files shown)


mikewiebe:
@saichint Did you run this across all nxos platforms?

saichint:
bridge domain is specific to n7k only. On other platforms, the CLIs "system bridge-domain ..." and "bridge-domain ..." do not exist. So the line I added is nop on other platforms. I verified on n9k just to be sure. It should be same for others. I don't have all the nxos platforms available right now to run but I don't think it matters.

on n7k --> this is the message we get when we run beaker (if bridge-domains exist)
--------                                                                                                                              
     * TestStep :: bridge-domain CLEANUP ::                                                                                               
        * bridge-domain CLEANUP ::  Removing cisco_bridge_domain '665'                                                                    
        * bridge-domain CLEANUP ::  Removing cisco_bridge_domain '666'                                                                    
        * bridge-domain CLEANUP ::  Removing cisco_bridge_domain '667'                                                                    
        * bridge-domain CLEANUP ::  Removing cisco_bridge_domain '668'                                                                    
    Setup switch for provider test :: PASS

on n9k --> this is the message we get when we run beaker:
--------                                                                                                                              
     * TestStep :: bridge-domain CLEANUP ::                                                                                               
    Setup switch for provider test :: PASS

PR #282, opened 2016-04-27
Feature/interface beaker fixes

robert-w-gries:
## Fixes
* Test `speed`, `duplex`, and `negotiate auto` properties
  * Use `interface_capabilities` to get valid `speed` and `duplex` values
* Replace hardcoded interface title patterns in `test_interface_stp.rb` with result of `find_interfaces`
* Fix `test_vlan_mapping.rb` by specifying proper switchport mode

## Changes
* Allow `tests[id]` to have its own specified `:resource_name`
* Ensure that `find_interfaces` does not user a subif (aka `Ethernet1/1.1`)
* Prevent duplicate `auto` values from being in `probe`


robert-w-gries:
All platforms have been tested, except for Fretta

.rubocop.yml
```diff
@@ -7,7 +7,7 @@ AllCops:
 
 # Baseline code complexity metrics - we should try and reduce these over time
 Metrics/AbcSize:
-  Max: 47
+  Max: 71.5
```
chrisvanheuveln: Ugh. I'd rather have this cop disabled only in the exceptions (affected files) rather than across the board.

tests/beaker_tests/cisco_interface/test_interface.rb
```diff
@@ -572,10 +586,18 @@ def test_harness_interface(tests, id)
...
-  logger.info("\n#{'-' * 60}\nSection 5. MISC Property Testing")
-  test_harness_interface(tests, 'negotiate')
-  # TBD: test_harness_interface(tests, 'speed_dup_mtu')
+  ### -------------------------------------------------------------------
+  logger.info("\n#{'-' * 60}\nSection 5. Negotiate Auto, MTU, Speed, Duplex")
+  if interface_pre_check(tests)
+    test_harness_run(tests, :auto)
+    test_harness_run(tests, :non_default)
+  else
+    msg = 'Could not find interface capabiilities'
```
chrisvanheuveln: `/iilities/ilities/`

tests/beaker_tests/cisco_interface/test_interface.rb
```diff
@@ -364,37 +364,28 @@
...
-tests['negotiate'] = {
-  desc:               '5.1 negotiate-auto',
+tests[:auto] = {
+  desc:               '5.1 Misc. Auto Value Properties',
   operating_system:   'nexus',
-  platform:           'n(5|6|7)k',
   intf_type:          'ethernet',
-  preclean:           true,
   sys_def_switchport: false,
+  resource_name:      'cisco_interface',
```
chrisvanheuveln: Why do you need `:resource_name` in the test case hash itself? What does this do that `tests[:resource_name]` wasn't already doing?

chrisvanheuveln:
👍 Relatively minor stuff to address. Good fixes overall.

.rubocop.yml
robert-w-gries: Why do you prefer disabling the cop?  imo it's easier to track the metric size in `.rubocop.yml` than tracking multiple cop disables.

tests/beaker_tests/cisco_interface/test_interface.rb
robert-w-gries: I was pretty sure I had a good reason, but testing has proven me wrong :-)

The `resource_name` changes will be removed in the next push.

tests/beaker_tests/cisco_interface/test_interface.rb
robert-w-gries: Fixed

.rubocop.yml
chrisvanheuveln: The size is the problem. If we keep letting this number grow then it becomes fairly meaningless and it won't flag methods that are getting out of hand. Some methods are just going to be large and it won't make sense to carve them up into smaller pieces, but those should be the exceptions. 

By disabling this on case-by-case basis, when we see a disabled around a particular method in review it's going to catch our attention and we can make a judgement call at that point.

I can't say for sure what the optimum size for this should be, but the methods it's been flagging lately are clearly too big so I would venture that the current value is pretty close to the max we want.

tests/beaker_tests/cisco_interface/test_interface.rb
chrisvanheuveln: Been there :-)

robert-w-gries:
Latest changes pushed

PR #289, opened 2016-04-28
Vxlan Puppet changes for release 1.3

anramgop:
- Included support for all nexus platforms (excluding 3k)
- Hold down time, ingress replication, peer list specific to 8,9k.
- added suppress_uuc - specific to 5,6k.
- changes in beaker tests accordingly.
- Manifest - ensure f3 vdc for 7k. Vxlan included in demo_all.

### DIFF
--- a/examples/cisco/demo_vxlan.pp
+++ b/examples/cisco/demo_vxlan.pp
@@ -17,80 +17,88 @@
 # limitations under the License.
 
 class ciscopuppet::cisco::demo_vxlan {
-  cisco_overlay_global { 'default':
-    dup_host_ip_addr_detection_host_moves => '100',
-    dup_host_ip_addr_detection_timeout    => '10',
-    anycast_gateway_mac                   => '1234.4567.6789',
-    dup_host_mac_detection_host_moves     => '100',
-    dup_host_mac_detection_timeout        => '10',
-  }
 
-  if platform_get() =~ /n7k/ {
-    cisco_vdc { 'default':
-      ensure                     => present,
-      limit_resource_module_type => 'f3'
+  if platform_get() =~ /n(5|6|7|8|9)k/ {
+
+    if platform_get() =~ /n7k/ {
+      cisco_vdc { 'default':
+        ensure                     => present,
+        limit_resource_module_type => 'f3'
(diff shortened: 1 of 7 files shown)


examples/cisco/demo_vxlan.pp
```diff
@@ -25,35 +25,72 @@
     dup_host_mac_detection_timeout        => '10',
   }
```
chrisvanheuveln: Since this isn't supported on 3k/XR you should have all of this code in an if block starting at line 20 and a notify/skip at the end of the file; e.g.:

```
  if platform_get() =~ /n(5|6|7|8|9)k/ {
    cisco_overlay_global { 'default':

       ...blah...

       ...blah...

  } else {
    notify{'SKIP: This platform does not support vxlan': }
  }

```

tests/beaker_tests/cisco_vxlan_vtep_vni/test_vxlan_vtep_vni.rb
```diff
@@ -353,6 +387,14 @@ def test_harness_cisco_vxlan_vtep_vni(tests, id)
   # tests[id][:desc] = '2.8 Suppress ARP'
   # test_harness_cisco_vxlan_vtep_vni(tests, id)
 
+  # id = 'suppress_uuc_true'
+  # tests[id][:desc] = '2.9 Suppress Unknown Unicast'
+  # test_harness_cisco_vxlan_vtep_vni(tests, id)
+  #
+  # id = 'suppress_uuc_false'
+  # tests[id][:desc] = '2.10 Suppress Unknown Unicast'
+  # test_harness_cisco_vxlan_vtep_vni(tests, id)
```
chrisvanheuveln: What's the plan for these commented out tests?

Also, just an fyi -- the minitests are responsible for testing every possible property state so we don't typically test every state in our beaker tests (unless there's a good reason to do so). We normally just do a default test (for properties that support default) and a non-default test (of each property). Additional tests are only done if we have incompatible properties that have to be tested separately.

chrisvanheuveln:
Just minor stuff to address. Rest looks good. 👍

PR #297, opened 2016-04-29
N7k testbed cleanup fixes

chrisvanheuveln:
* Mostly F3 & vlan related
* The F3 issue still needs work but we will address post-release.
* feature vni can only work if limit-resource module-type is set exclusively to 'f3'
* Tested (on N7k w/F3 module) all of the tests that use the methods I changed

### DIFF
--- a/tests/beaker_tests/cisco_fabricpath_global/test_fabricpath_global.rb
+++ b/tests/beaker_tests/cisco_fabricpath_global/test_fabricpath_global.rb
@@ -238,14 +238,16 @@ def test_harness_fabricpath_global(tests, id)
   tests[id][:ensure] = nil
 end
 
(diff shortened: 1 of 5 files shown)


tests/beaker_tests/cisco_fabricpath_global/test_fabricpath_global.rb
```diff
@@ -238,14 +238,16 @@ def test_harness_fabricpath_global(tests, id)
   tests[id][:ensure] = nil
 end
 
+def testbed_cleanup(agent)
+  remove_all_vlans(agent)
```
mikewiebe: Does this cleanup work on n5k and n6k since fabric path is supported on those platforms?

PR #301, opened 2016-04-30
Bridge-Domain fix for beaker tests

dcheriancisco:


tests/beaker_tests/cisco_stp_global/test_stp_global.rb
```diff
@@ -280,6 +286,8 @@ def test_harness_dependencies(_tests, id)
   test_harness_run(tests, :non_default_bd)
   test_harness_run(tests, :non_default_plat_1)
   test_harness_run(tests, :non_default_plat_2)
+
+  test_cleanup_all_bridge_domains
```
chrisvanheuveln: Should this test also do a vlan/bd cleanup prior to the `test_harness_run` calls?

tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -1050,6 +1050,19 @@ def find_interface(tests, id=nil, skipcheck=true)
     msg = 'Unable to find suitable interface module for this test.'
     prereq_skip(tests[:resource_name], self, msg)
   end
+
+  if type =~ /ethernet/i
```
chrisvanheuveln: * This block should be part of the ethernet case block above. 
* This block should only process for modular switches, `if platform[/n7k/]`
* Tests that use feature vni will fail with 'f2e f3', must use 'f3' only
* We have other methods already that try to handle f3 setup: `setup_mt_full_env`, `setup_fabricpath_env`, etc. How will this change play with those methods? I think we may want to take a step back and come up with a comprehensive setup instead.

chrisvanheuveln:
Discussed with @mikewiebe  -- we want to include the bd cleanup for 1.3.0 but handle the F3 setup post-release. We need to merge to master by noon today so I'll go ahead and commit the bd part myself.

tests/beaker_tests/lib/utilitylib.rb
dcheriancisco: The idea is that I want to work on an interface already selected by the first block and *ONLY* if that selected interface is on an N7K/F2 card (see the check on line 1058). if the other setup functions have caused a limit-resource to be with 'f3' cards, it would fail the guard below. So there should be no interaction with other setup methods.
The issue is specific to f2/f2e cards, so every L3 feature which chooses an f2/f2e interface would need to change. I thought this is a common function used by such features. So a common place to change.

tests/beaker_tests/cisco_stp_global/test_stp_global.rb
dcheriancisco: yes it does that in line 268 already.  Probably there should be a guard for n7k product-id for this function.

tests/beaker_tests/lib/utilitylib.rb
chrisvanheuveln: We have a meeting scheduled after scrum to discuss a long-term fix for testbed cleanup. I'll send you the invite.

chrisvanheuveln:
Closing for now. US104167 will address this work. I've included a link to this PR in that User Story.

PR #314, opened 2016-05-05
Refactor test_interface.rb into multiple files

chrisvanheuveln:
* This test was trying to do too much; it became too large and unwieldy to use and troubleshoot.
* Refactor to use new test syntax (test_harness_run)
* File changes:
  *  deleted:    test_interface.rb
  *  new file:   interfacelib.rb
  *  new file:   test_interface_L2.rb
  *  new file:   test_interface_L3.rb
  *  new file:   test_interface_bdi.rb
  *  modified:   test_interface_capabilities.rb
  *  modified:   test_interface_stp.rb
  *  new file:   test_interface_svi.rb
  *  new file:   test_interface_private_vlan.rb

* renamed test_vlan_mapping.rb to test_interface_vlan_mapping.rb

* Only tested on n7k and n9k so far but there's a lot here and I wanted to get some eyes on the changes


tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -357,6 +323,19 @@ def system_default_switchport(agent, state=false,
   end
 end
 
+# Helper for checking/setting 'system default switchport'
+def config_system_default_switchport?(tests, id)
```
mikewiebe: This is find, but did you consider combining `config_system_default_switchport?` and `config_system_default_switchport_shutdown?` since the code is identical but add a `key` argument?

`def config_system_defaults?(tests, id, key)`

tests/beaker_tests/cisco_interface/interfacelib.rb
```diff
@@ -0,0 +1,66 @@
...
+  if intf[/ethernet/i]
+    interface_cleanup(agent, intf)
+  else
+    remove_interface(agent, intf)
+  end
+end
+
+# A global 'anycast gateway mac' is required for some SVI properties
+# TBD: This may not be required for all platforms
+def config_anycast_gateway_mac?(tests, _id)
```
mikewiebe: Is this something that would be useful in `utilitylib.rb`?

tests/beaker_tests/cisco_interface/interfacelib.rb
```diff
@@ -0,0 +1,66 @@
...
+  return unless tests.key?(:anycast_gateway_mac)
+  agent = tests[:agent]
+  on(agent, get_vshell_cmd('show runn fabric forwarding'), pty: true)
```
mikewiebe: Instead of adding another `vshell` dependency, what do you think about creating a `resource_get` method to query the device using the `puppet resource` command?  We have a `resource_set` but a `resource_get` would be a nice compliment.

tests/beaker_tests/lib/utilitylib.rb
chrisvanheuveln: I did. They're not identical, they do the actual config change while my new methods are just used to check if the changes are actually necessary. Also, some callers are calling those direct, in my case I'm just front-ending them to avoid having to set the state for every test case.

I resisted overloading the existing methods but I can add my logic to them if you think that would be cleaner.

tests/beaker_tests/cisco_interface/interfacelib.rb
chrisvanheuveln: I considered it but I didn't see a use for it outside the interface tests. If we find a use case later we can move it to utilitylib.

tests/beaker_tests/cisco_interface/interfacelib.rb
chrisvanheuveln: I agree but I don't want to add it to this PR.

Note that a trade-off with resource_get will be slower tests (puppet resource is painfully slow).

mikewiebe: True, but we need to move away from vsh as much as possible.  Just add a TBD then.

tests/beaker_tests/cisco_interface/test_interface_L2.rb
```diff
@@ -0,0 +1,127 @@
...
+tests[:default_access] = {
+  desc:               "1.1 Default 'access' Properties",
+  title_pattern:      intf,
+  code:               [0],
+  preclean_intf:      true,
+  sys_def_switchport: true,
+  sys_def_sw_shut:    true,
+  manifest_props:     {
+    shutdown:        'default',
+    switchport_mode: 'access',
```
mikewiebe: Remind me why we don't have a `switchport_mode` test for default?

https://github.com/cisco/cisco-network-puppet-module/blob/develop/lib/puppet/type/cisco_interface.rb#L212

tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -1211,6 +1226,12 @@ def debug_probe(probe, msg)
   logger.info("\n      #{msg}: #{dbg}")
 end
 
+# Remove a single dynamic interface (Vlan, Loopback, Port-channel, etc).
+def remove_interface(agent, intf)
+  cmd = "    no interface #{intf.capitalize}"
```
mikewiebe: Is the whitespace needed?

tests/beaker_tests/cisco_interface/test_interface_L2.rb
chrisvanheuveln: We probably left it out because we have to first set the mode to access, trunk, etc before we can test the associated properties for that mode. It might be worthwhile to add a default test to this at some point.

tests/beaker_tests/lib/utilitylib.rb
chrisvanheuveln: Since it's indented it displays a little nicer when it's running.

PR #324, opened 2016-05-18
cisco_interface: private_vlan refactor

chrisvanheuveln:
* Deprecated the private_vlan properties that were released with 1.3.0 and created new pvlan properties in their place.

* This is the companion to NU PR: cisco/cisco-network-node-utils#406

* Updated Type, Provider, Beaker, Demo, README, and CHANGELOG.

* Currently only tested on N7, N9-i3. Additional testing in prog.

### DIFF
--- a/examples/cisco/demo_interface.pp
+++ b/examples/cisco/demo_interface.pp
@@ -108,24 +108,39 @@ class ciscopuppet::cisco::demo_interface {
       warning('This platform does not support cisco_bridge_domain')
     }
 
-    # For private vlan
+    # Private-vlan
     if platform_get() =~ /n(3|5|6|7|9)k/ {
       cisco_vlan { '445':
-        ensure     => present,
-        private_vlan_type => 'isolated',
+        ensure                              => present,
+        #pvlan_type                          => 'isolated',
+        private_vlan_type => 'isolated',     # DEPRECATED
       }
       cisco_vlan { '444':
-        ensure     => present,
-        private_vlan_type => 'primary',
-        private_vlan_association => ['445'],
+        ensure                              => present,
+        #pvlan_type                          => 'primary',
+        #pvlan_association                   => ['445'],
+        private_vlan_type  => 'primary',     # DEPRECATED
+        private_vlan_association => ['445'], # DEPRECATED
       }
       cisco_interface { 'Ethernet1/6':
-        description     => 'private_vlan_host_port',
-        switchport_mode_private_vlan_host => 'host',
-        switchport_mode_private_vlan_host_association  => ['444', '445'],
+        description                         => 'Private-vlan Host Port',
+        switchport_pvlan_host               => true,
+        switchport_pvlan_host_association   => [444, 445],
+      }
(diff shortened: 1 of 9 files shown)


examples/cisco/demo_interface.pp
```diff
@@ -108,24 +108,39 @@
...
 
-    # For private vlan
+    # Private-vlan
     if platform_get() =~ /n(3|5|6|7|9)k/ {
       cisco_vlan { '445':
-        ensure     => present,
-        private_vlan_type => 'isolated',
+        ensure                              => present,
+        #pvlan_type                          => 'isolated',
+        private_vlan_type => 'isolated',     # DEPRECATED
```
mikewiebe: Why are the non-deprecated properties commented out?

chrisvanheuveln: I was waiting for Sai's commit #325. I just merged that so I'll update this now.

lib/puppet/type/cisco_interface.rb
```diff
@@ -728,100 +702,268 @@ def is_to_s(value)
...
-  newproperty(:switchport_mode_private_vlan_host) do
-    desc 'Switchport private host mode of the interface.'
+  # ------------------------
+  newproperty(:pvlan_mapping, array_matching: :all) do
+    inputs =
+      'Valid inputs are a String containing a range of secondary vlans, '\
+      "example: '3-4,6'; or keyword 'default'"
+    desc 'Maps secondary VLANs to the VLAN interface of a primary VLAN. ' + inputs
+
+    validate do |value|
```
mikewiebe: Do we need to be concerned about an invalid manifest entry like this? `'5-2, 1'`

chrisvanheuveln: No need to worry. That's not invalid -- the `merge_ranges` code will merge and sort the ranges correctly.

tests/beaker_tests/cisco_interface/test_interface_private_vlan.rb
```diff
@@ -51,114 +51,91 @@
...
-    switchport_mode_private_vlan_host:              'disabled',
-    # switchport_mode_private_vlan_host_association: nil,
-    switchport_mode_private_vlan_trunk_promiscuous: 'false',
-    switchport_mode_private_vlan_trunk_secondary:   'false',
-    # switchport_private_vlan_trunk_allowed_vlan:    nil,
-    # switchport_private_vlan_association_trunk:     nil,
-    # switchport_private_vlan_mapping_trunk:         nil,
-    switchport_private_vlan_trunk_native_vlan:      '1',
+    switchport_pvlan_host:               'false',
+    # switchport_pvlan_host_association:  nil,
```
mikewiebe: Uncomment?

chrisvanheuveln: No - these nil properties don't show up with puppet resource. I thought it would be helpful to put them here as comments so that future coders would understand why they're not part of the resource hash.

PR #328, opened 2016-05-24
Refactor package provider to use rpm - qip

mikewiebe:
**Summary:**
Our existing package provider made use of fairly slick regular expressions to decompose the rpm filename and use the various bits of information to ensure idempotency.  The problem is that the regular expressions did not work in all cases since rpm filenames can be quite diverse.  This update instead uses the `rpm -qip` command to grab the rpm metadata and use that to ensure idempotency.

**Tested On:**
- n9k(I2), n9k(I4), n8k


lib/puppet/provider/package/cisco.rb
```diff
@@ -77,41 +77,53 @@ def properties
...
+    r_re = /Release(?:\s+ )?:\s+(\S+)/
+    a_re = /Architecture(?:\s+ )?:\s+(\S+)/
+    name = n_re.match(rpm_data) ? Regexp.last_match(1) : nil
+    ver  = v_re.match(rpm_data) ? Regexp.last_match(1) : nil
+    rel  = r_re.match(rpm_data) ? Regexp.last_match(1) : nil
+    arch = a_re.match(rpm_data) ? Regexp.last_match(1) : nil
+
+    [name, ver, rel, arch].each do |val|
+      fail "Unable parse rpm data from #{resource[:source]}\n#{rpm_data}" \
+        if val.nil?
```
chrisvanheuveln: nit: another way:
```
-    [name, ver, rel, arch].each do |val|
-      fail "Unable parse rpm data from #{resource[:source]}\n#{rpm_data}" \
-        if val.nil?
-    end
+    fail "Unable to parse rpm data from #{resource[:source]}\n#{rpm_data}" if
+      [name, ver, rel, arch].include?(nil)
```

PR #333, opened 2016-05-31
Misc test_interface fixes

chrisvanheuveln:
* test_interface_L2 should reset sys_def_switchport
  * This test sets the system default switchport state to true but it should reset it to false when it's finished.

* Remove explicit anycast config
  * anycast-gateway-mac is a required dependency; a recent commit added an explicit call to the test file to set this up; BUT we already had existing code for doing this that was part of interfacelib.rb

  * The interfacelib code was returning early because of /unless/if/

* Tested on n9-i4


chrisvanheuveln:
This should have been a direct commit. I'll merge it myself.

PR #336, opened 2016-06-02
test_vdc: fix for non-F3 testbeds

chrisvanheuveln:
* Updated this test to work with non-F3 N7k testbeds.
  * Initially set the VDC module-type to default state
  * Then set module-type to F3 and test
  * There's no great way to know which way to leave the module-type when this test completes, so:
    * Reset module-type to default for non-F3 testbeds since leaving it as F3 means the device won't have any usable interfaces
    * Noop for F3 testbeds so that it leaves their F3 still usable

* Tested on n6, n7-non-F3

Edit: Added some add'l fixes for bridge-domain tests

### DIFF
--- a/tests/beaker_tests/cisco_bridge_domain/test_bridge_domain.rb
+++ b/tests/beaker_tests/cisco_bridge_domain/test_bridge_domain.rb
@@ -29,6 +29,9 @@ tests = {
   platform:      'n7k',
 }
 
+# Skip -ALL- tests if a top-level platform/os key exludes this platform
+skip_unless_supported(tests)
+
 tests[:non_default] = {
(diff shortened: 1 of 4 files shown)


chrisvanheuveln:
Would it make sense for this PR to merge with develop instead of release_1.3.2? I don't think it's that risky (it only affects n7k) but it's not particulary pressing for 1.3.2 either.

mikewiebe:
@chrisvanheuveln The decision to put this into release_1.3.2 will depend on the following.
- How many other issues does this uncover in other test areas that were getting skipped before the change.
- How quickly can we fix the remaining issues in advance of the 6/7 release date.

PR #340, opened 2016-06-03
bgp_neighbor bfd

saichint:
This PR is for extending bgp_neighbor provider with bfd property.
All beaker and demo_manifest tests have passed on all nexus platforms.


chrisvanheuveln:
Nice work Sai.

saichint:
Thank you very much. This is my first merge without any comments (though it is a small PR).

PR #347, opened 2016-06-09
Test get set

chrisvanheuveln:
Add test_get / test_set to command_config
* Needed a way for beaker to query / set config that didn't have explicit resource support (e.g. disabling a feature)

* test_get is a simple get-only property that does a show running; callers set a filter to parse what they need:
    puppet resource cisco_command_config 'cc' test_get='incl feature'

* test_set is a simple set-only property:
    puppet resource cisco_command_config 'cc' test_set='no feature foo'

* Also fixed a bad regex in file_service_package

* Tested on n3,n5,n7,n9


chrisvanheuveln:
This needs to go into develop instead of 1.3.2 so I'll reopen for that.

PR #348, opened 2016-06-09
Add test_get / test_set to command_config

chrisvanheuveln:
- Needed a way for beaker to query / set config that didn't have explicit resource support (e.g. disabling a feature)
- test_get is a simple get-only property that does a show running; callers set a filter to parse what they need:
      puppet resource cisco_command_config 'cc' test_get='incl feature'
- test_set is a simple set-only property:
      puppet resource cisco_command_config 'cc' test_set='no feature foo'
- Tested on n3,n5,n7,n9
- edit: Updated this PR to include some fixes I did for the fabricpath tests:
  - Added new utility: config_find_remove() uses test_get to search for existing configs, then test_set to remove it.
  - Fixed the fabricpath tests which ran into an issue on n5k when cleaning up 'feature nv overlay' with command_config: Puppet resource removes the config but then immediately does a 'show runn' which gets a SAP timeout on the device itself and the test errors. Using the test_set setter instead of regular command_config gets around this problem because it's a "set-only" with a dummy test_set getter.
  - Latest code tested on n5,n7,n9

### DIFF
--- a/lib/puppet/provider/cisco_command_config/cisco.rb
+++ b/lib/puppet/provider/cisco_command_config/cisco.rb
@@ -95,16 +95,21 @@ Puppet::Type.type(:cisco_command_config).provide(:cisco) do
     cmd = 'show running-config all'
     cmd << " | #{@resource[:test_get]}" if @resource[:test_get]
 
-    result = @node.get(command: cmd)
-    "\n" + result unless result.nil?
+    output = @node.get(command: cmd)
+    debug "@node.get output:\n#{output}"
+    "\n" + output unless output.nil?
   end
 
   def test_get=(noop)
-    # This is a dummy setter to keep Puppet from raising an error.
+    # This is a dummy "setter" for test_get(), which is a get-only property.
+    # This dummy method is necessary to keep Puppet from raising an error or
+    # displaying noise.
   end
 
   def test_set
-    # This is a dummy getter to keep Puppet from raising an error.
+    # This is a dummy "getter" for test_set=(), which is a set-only property.
+    # This dummy method is necessary to keep Puppet from raising an error or
+    # displaying noise.
   end
 
   def test_set=(cmds)
@@ -114,6 +119,6 @@ Puppet::Type.type(:cisco_command_config).provide(:cisco) do
     #  puppet resource cisco_command_config 'cc' test_set='no feature foo'
     return if cmds.empty?
     output = @node.set(values: cmds)
-    debug "Output from node:\n#{output}" unless output.nil?
+    debug "@node.set output:\n#{output}" unless output.nil?
   end
 end # Puppet::Type


lib/puppet/provider/cisco_command_config/cisco.rb
```diff
@@ -85,4 +85,35 @@ def command=(cmds)
...
+    # Example usage:
+    #  puppet resource cisco_command_config 'cc' test_get='incl feature'
+    cmd = 'show running-config all'
+    cmd << " | #{@resource[:test_get]}" if @resource[:test_get]
+
+    result = @node.get(command: cmd)
+    "\n" + result unless result.nil?
+  end
+
+  def test_get=(noop)
```
mikewiebe: your set and get method names are reversed :)

lib/puppet/provider/cisco_command_config/cisco.rb
```diff
@@ -85,4 +85,35 @@ def command=(cmds)
...
+    # This method is for beaker use only. It allows beaker to retrieve any
+    # configuration it needs from the device using puppet resource. Callers
+    # must pass a filter string to test_get.
+    # Example usage:
+    #  puppet resource cisco_command_config 'cc' test_get='incl feature'
+    cmd = 'show running-config all'
+    cmd << " | #{@resource[:test_get]}" if @resource[:test_get]
+
+    result = @node.get(command: cmd)
+    "\n" + result unless result.nil?
```
mikewiebe: I think it would be helpful to add a debug here as well to dump out `result`.  Useful for troubleshooting.

lib/puppet/provider/cisco_command_config/cisco.rb
```diff
@@ -85,4 +85,35 @@ def command=(cmds)
...
+  end
+
+  def test_set=(cmds)
+    # This method is for beaker use only. It allows beaker to set simple raw
+    # configuration using puppet resource.
+    # Example usage:
+    #  puppet resource cisco_command_config 'cc' test_set='no feature foo'
+    return if cmds.empty?
+    output = @node.set(values: cmds)
+    debug "Output from node:\n#{output}" unless output.nil?
```
mikewiebe: **Nit** `debug "Output from @node.set command:\n#{output}" unless output.nil?`

lib/puppet/type/cisco_command_config.rb
```diff
@@ -106,4 +106,37 @@ def self.title_patterns
...
+      must pass a filter string to test_get.
+      Example usage:
+        puppet resource cisco_command_config 'cc' test_get='incl feature'
+    )
+
+    def insync?(*)
+      # This is a "get-only" property so insync? is overridden to prevent
+      # puppet from displaying the following notice:
+      #   Notice: /Cisco_command_config[c]/test_get: test_get changed '' to ''
+      true
```
mikewiebe: Interesting.  Would love to know how you figured this out! :)

tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -800,6 +800,48 @@ def setup_fabricpath_env(tests, testcase)
...
+end
+
+# Get raw configuration from the device using command_config's test_get.
+# test_get does a 'show runn' but requires a filter.
+# Example:
+#  test_get(agent, 'incl ^vlan')
+def test_get(agent, filter)
+  cmd_prefix = PUPPET_BINPATH + "resource cisco_command_config 'cc' "
+  on(agent, cmd_prefix + "test_get='#{filter}'")
+  stdout
```
mikewiebe: Does this get cleared on every `on` call?

lib/puppet/provider/cisco_command_config/cisco.rb
chrisvanheuveln: added

lib/puppet/provider/cisco_command_config/cisco.rb
chrisvanheuveln: Nope :-)

Added add'l comments to clear the confusion.

lib/puppet/provider/cisco_command_config/cisco.rb
chrisvanheuveln: done

lib/puppet/type/cisco_command_config.rb
chrisvanheuveln: It's the idempotence checker method so I just deduced that always returning true would avoid the Notice.

tests/beaker_tests/lib/utilitylib.rb
chrisvanheuveln: yep

PR #352, opened 2016-06-13
Ospf_area provider

saichint:
This PR is for cisco_ospf_area provider. The properties do not include nssa and its associated properties. They will be added as part of another user story.

Since this provider becomes absent when all properties are set to default, cisco_pim model is followed for beaker tests.

All demo_manifests and beaker tests passed on all platforms.


README.md
```diff
@@ -2403,6 +2406,52 @@ and 'absent'.
...
+Determines whether the config should be present or not on the device. Valid values are 'present' and 'absent'.
+
+##### `authentication`
+Enables authentication for the area. Valid values are 'clear_text', 'md5' or 'default'.
+
+##### `default_cost`
+Default_cost for default summary Link-State Advertisement (LSA). Valid values are integer or keyword 'default'.
+
+##### `filter_list_in`
+Filter networks sent to this area. Valid values are string or keyword 'default'.
```
```
This is a route-map for filtering networks sent to this area. Valid values are string or keyword 'default'.
```

README.md
```diff
@@ -2403,6 +2406,52 @@ and 'absent'.
...
+
+##### `filter_list_out`
+Filter networks sent from this area. Valid values are string or keyword 'default'.
```
chrisvanheuveln: ditto similar comment as above

README.md
```diff
@@ -2403,6 +2406,52 @@ and 'absent'.
...
+
+##### `range`
+Configures an address range for an area.  Valid values are an array of [ip, not_advertise, cost] pairs or keyword 'default'. `ip` is mandatory in the sub-arrays but `not_advertise` and `cost` are optional.
```
chrisvanheuveln: Range is really about creating summary routes, so how about something like this:

```
Summarizes routes at an area boundary. Optionally sets the area range status to DoNotAdvertise as well as per-summary cost values. Valid values are a nested array of [summary_address, 'not_advertise', cost], or keyword 'default'. The summary-address is mandatory.
```

README.md
```diff
@@ -2403,6 +2406,52 @@ and 'absent'.
...
+                    ['10.3.0.0/32', 'not_advertise'],
+                    ['10.3.0.1/32'],
+                    ['10.3.3.0/24', '450']]`
+
+##### `stub`
+Configures the area as a stub. Valid values are true, false or keyword 'default'.
+
+
+##### `stub_no_summary`
+Prevents Area Border Router (ABR) from sending summary LSAs into stub area. Valid values are true, false or keyword 'default'.
```
```
Stub areas flood summary LSAs. This property disables summary flooding into the area. This property can be used in place of the `stub` property or in conjunction with it. Valid values are true, false or keyword 'default'.
```

README.md
```diff
@@ -2403,6 +2406,52 @@ and 'absent'.
...
+##### `range`
+Configures an address range for an area.  Valid values are an array of [ip, not_advertise, cost] pairs or keyword 'default'. `ip` is mandatory in the sub-arrays but `not_advertise` and `cost` are optional.
+
+Example: `range => [['10.3.0.0/16', 'not_advertise', '23'],
+                    ['10.3.0.0/32', 'not_advertise'],
+                    ['10.3.0.1/32'],
+                    ['10.3.3.0/24', '450']]`
+
+##### `stub`
+Configures the area as a stub. Valid values are true, false or keyword 'default'.
```
```
Defines the area as a stub area. Valid values are true, false or keyword 'default'. This property is not necessary when the `stub_no_summary` property is used, which also defines the area as a stub area.
```

examples/cisco/demo_ospf.pp
```diff
@@ -64,4 +64,28 @@
...
+
+  cisco_ospf_area { 'dark_blue default 1.1.1.1':
+    require         => Cisco_ospf_vrf['dark_blue default'],
+    ensure          => 'present',
+    authentication  => 'md5',
+    default_cost    => 1000,
+    filter_list_in  => 'fin',
+    filter_list_out => 'fout',
+    range           => [['10.3.0.0/16', 'not_advertise', '23'], ['10.3.3.0/24', '450']],
+    stub            => true,
```
chrisvanheuveln: I would remove this line

README.md
```diff
@@ -2403,6 +2406,52 @@ and 'absent'.
...
+
+| Platform | OS Minimum Version | Module Minimum Version |
+|----------|:------------------:|:----------------------:|
+| N9k      | 7.0(3)I3(1)        | 1.4.0                  |
+| N3k      | 7.0(3)I3(1)        | 1.4.0                  |
+| N5k      | 7.3(0)N1(1)        | 1.4.0                  |
+| N6k      | 7.3(0)N1(1)        | 1.4.0                  |
+| N7k      | 7.3(0)D1(1)        | 1.4.0                  |
+| N8k      | 7.3(0)F1(1)        | 1.4.0                  |
+
```
chrisvanheuveln: I would like to see some example usage documentation also. This is a gap in our docs. I did this for ACE: https://github.com/saichint/cisco-network-puppet-module/blob/a67dd1c7efc2bf014bedee5102092e8ca9736c52/README.md#example-usage

So, maybe just some basic examples like this:

```
#### Example Usage

cisco_ospf_area { 'my_ospf_instance  default  0':
  ensure          => 'present',
  range           => [['10.3.0.0/16', 'not_advertise', '23'],
                            ['10.3.3.0/24', '450']
                           ],
}

cisco_ospf_area { 'my_ospf_instance  my_vrf  1.1.1.1':
  ensure          => 'present',
  authentication  => 'md5',
  default_cost    => 1000,
  filter_list_in  => 'fin',
  filter_list_out => 'fout',
  stub             => true,
}

#### Parameters


| Example Parameter Usage |
|:--|:--
|`cisco_ospf_area { '<ospf_process_id> <vrf> <area_id>':`
|`cisco_ospf_area { '1 my_vrf 0':`
|`cisco_ospf_area { 'my_ospf default 10.1.1.1':`

```

lib/puppet/provider/cisco_ospf_area/cisco.rb
```diff
@@ -0,0 +1,168 @@
...
+    debug "Checking ospf instance, #{ospf} #{vrf} #{area}"
+    current_state = {
+      name:   "#{ospf} #{vrf} #{area}",
+      ospf:   ospf,
+      vrf:    vrf,
+      area:   area,
+      ensure: :present,
+    }
+    # Call node_utils getter for each property
+    OSPF_AREA_NON_BOOL_PROPS.each do |prop|
```
chrisvanheuveln: nit: You can handle the array here too, ie:

```
(OSPF_AREA_NON_BOOL_PROPS + OSPF_AREA_ARRAY_FLAT_PROPS).each do |prop|
  current_state[prop] = nu_obj.send(prop)
end

File: lib/puppet/provider/cisco_ospf_area/cisco.rb [hunk 9]
```diff
@@ -0,0 +1,168 @@
...
+    :default_cost,
+    :filter_list_in,
+    :filter_list_out,
+  ]
+  OSPF_AREA_BOOL_PROPS = [
+    :stub,
+    :stub_no_summary,
+  ]
+  OSPF_AREA_ARRAY_FLAT_PROPS = [
+    :range
```
chrisvanheuveln: You should be using the nested array autogen instead.

lib/puppet/provider/cisco_ospf_area/cisco.rb
```diff
@@ -0,0 +1,168 @@
...
+
+  OSPF_AREA_ALL_PROPS = OSPF_AREA_NON_BOOL_PROPS + OSPF_AREA_BOOL_PROPS +
+                        OSPF_AREA_ARRAY_FLAT_PROPS
+
+  PuppetX::Cisco::AutoGen.mk_puppet_methods(:non_bool, self, '@nu',
+                                            OSPF_AREA_NON_BOOL_PROPS)
+  PuppetX::Cisco::AutoGen.mk_puppet_methods(:bool, self, '@nu',
+                                            OSPF_AREA_BOOL_PROPS)
+  PuppetX::Cisco::AutoGen.mk_puppet_methods(:array_flat, self, '@nu',
+                                            OSPF_AREA_ARRAY_FLAT_PROPS)
```
chrisvanheuveln: `/FLAT/NESTED/`

lib/puppet/provider/cisco_ospf_area/cisco.rb
```diff
@@ -0,0 +1,168 @@
...
+    end
+    OSPF_AREA_BOOL_PROPS.each do |prop|
+      val = nu_obj.send(prop)
+      if val.nil?
+        current_state[prop] = nil
+      else
+        current_state[prop] = val ? :true : :false
+      end
+    end
+    OSPF_AREA_ARRAY_FLAT_PROPS.each do |prop|
```
chrisvanheuveln: Remove. See above.

lib/puppet/provider/cisco_ospf_area/cisco.rb
```diff
@@ -0,0 +1,168 @@
...
+      current_state[prop] = nu_obj.send(prop)
+    end
+    # nested array properties
+    current_state[:range] = nu_obj.range
```
chrisvanheuveln: This call is not needed when you use the nested methods.

lib/puppet/provider/cisco_ospf_area/cisco.rb
```diff
@@ -0,0 +1,168 @@
...
+    new(current_state)
+  end # self.properties_get
+
+  def self.instances
+    area_instances = []
+    Cisco::RouterOspfArea.areas.each do |ospf, vrf|
+      vrf.each do |name, areas|
+        areas.each do |area|
+          area_instances << properties_get(ospf, name, area[0], area[1])
```
chrisvanheuveln: This code block really confuses me. 'name' is rarely a helpful variable name when you have several objects to deal with. Consider this instead:

```
   Cisco::RouterOspfArea.areas.each do |ospf, vrfs|
     vrfs.each do |vrf, areas|
     areas.each do |area, nu_obj|
       area_instances << properties_get(ospf, vrf, area, nu_obj)
```

lib/puppet/provider/cisco_ospf_area/cisco.rb
```diff
@@ -0,0 +1,168 @@
...
+      next unless @resource[prop]
+      send("#{prop}=", @resource[prop]) if new_area
+      unless @property_flush[prop].nil?
+        @nu.send("#{prop}=", @property_flush[prop]) if
+          @nu.respond_to?("#{prop}=")
+      end
+    end
+  end
+
+  def range=(should_list)
```
chrisvanheuveln: You can remove this after you switch to using the nested method.

lib/puppet/type/cisco_ospf_area.rb
```diff
@@ -0,0 +1,176 @@
...
+
+  newproperty(:default_cost) do
+    desc "default_cost for default summary Link-State Advertisement (LSA).
+          Valid values are integer, keyword 'default'."
+
+    munge { |value| value == 'default' ? :default : Integer(value) }
+  end # property default_cost
+
+  newproperty(:filter_list_in) do
+    desc "Filter networks sent to this area. Valid values are string,
```
chrisvanheuveln: Please update the `desc` strings to match the readme comments.

lib/puppet/type/cisco_ospf_area.rb
```diff
@@ -0,0 +1,176 @@
...
+         "Valid values match format #{format}."
+
+    # Override puppet's insync method, which checks whether current value is
+    # equal to value specified in manifest.  Make sure puppet considers
+    # 2 arrays with same elements but in different order as equal.
+    def insync?(is)
+      (is.size == should.size && is.sort == should.sort)
+    end
+
+    def should_to_s(value)
```
chrisvanheuveln: Please add a comment to explain why these are overridden. You told me once and it made sense but I can't remember what problem this solves. :-\

lib/puppet/type/cisco_ospf_area.rb
```diff
@@ -0,0 +1,176 @@
...
+    end
+
+    def is_to_s(value)
+      value.inspect
+    end
+
+    munge do |value|
+      begin
+        return value = :default if value == 'default'
+        fail("Value must match format #{format}") unless value.is_a?(Array)
```
chrisvanheuveln: Nit: validation should really be done by `validate`, not `munge`. It's okay here though.

tests/beaker_tests/cisco_ospf_area/test_ospf_area.rb
```diff
@@ -0,0 +1,100 @@
...
+
+# Test hash top-level keys
+tests = {
+  master:           master,
+  agent:            agent,
+  operating_system: 'nexus',
+  resource_name:    'cisco_ospf_area',
+}
+
+# Test hash test cases
```
chrisvanheuveln: No `:default` test case?

tests/beaker_tests/cisco_ospf_area/test_ospf_area.rb
```diff
@@ -0,0 +1,100 @@
...
+    filter_list_out: 'fout',
+    range:           rarray2,
+    stub:            'true',
+    stub_no_summary: 'false',
+  },
+  resource:       {
+    range: "#{rarray2}"
+  },
+}
+
```
chrisvanheuveln: Have you tested with an integer-based ospf ID? You don't need to add a formal test for it, just wondering if it works or not (it should).

tests/beaker_tests/cisco_ospf_area/test_ospf_area.rb
```diff
@@ -0,0 +1,100 @@
...
+# Overridden to properly handle dependencies for this test file.
+def test_harness_dependencies(_tests, _id)
+  cmd = 'feature ospf'
```
chrisvanheuveln: You only need to do this for the first test case so do this:

```
def test_harness_dependencies(_tests, id)
  return unless id == :default
```

This raises another question for me: Should the NU code do a feature enable for ospf? I'm thinking there's a use case for a very simple manifest for spoke routers, where all they want to do is make it part of the stub area so all they would have to do is the following, without having to  define cisco_ospf and cisco_ospf_vrf:

```
cisco_ospf_area { '1 default 1':
  stub => true
}
```

tests/beaker_tests/cisco_ospf_area/test_ospf_area.rb
```diff
@@ -0,0 +1,100 @@
...
+  device = platform
+  logger.info("#### This device is of type: #{device} #####")
+  logger.info("\n#{'-' * 60}\nSection 1.1 Non Default Property Testing")
+
+  test_harness_run(tests, :non_default_1)
+  # -------------------------------------------------------------------
+  logger.info("\n#{'-' * 60}\nSection 1.2 Non Default Property Testing")
+
+  test_harness_run(tests, :non_default_2)
+  resource_absent_cleanup(agent, 'cisco_ospf_area')
```
chrisvanheuveln: I would use `cisco_ospf` here instead since this test will leave an ospf instance behind otherwise.

tests/beaker_tests/cisco_ospf_area/test_ospf_area.rb
```diff
@@ -0,0 +1,100 @@
...
+# Test hash test cases
+
+rarray1 = Array[['10.3.0.0/16', 'not_advertise', '23'], ['10.3.3.0/24', '450']]
+rarray2 = Array[['10.3.0.0/16', '4989'], ['10.3.1.1/32']]
+
+# Non-default Tests. NOTE: [:resource] = [:manifest_props] for all non-default
+tests[:non_default_1] = {
+  desc:           '1.1 Non_Defaults',
+  title_pattern:  'dark_blue default 1.1.1.1',
+  preclean:       'cisco_ospf_area',
```
chrisvanheuveln: Use `cisco_ospf` to remove the entire instance.

tests/beaker_tests/cisco_ospf_area/test_ospf_area.rb
```diff
@@ -0,0 +1,100 @@
...
+  },
+  resource:       {
+    range: "#{rarray1}"
+  },
+}
+
+tests[:non_default_2] = {
+  desc:           '1.2 Non_Defaults',
+  title_pattern:  'dark_blue vrf1 2.2.2.2',
+  preclean:       'cisco_ospf_area',
```
chrisvanheuveln: I'd remove the preclean so that this is an additive test. Or just remove this testcase altogether. It's okay but I'm not sure it's adds that much value.

README.md
saichint: This one is a bit problematic. The way I implemented is that, stub property MUST be set to true for the stub_no_summary to take affect. So we can have stub as true and stub_no_summary as true or false. Also stub can be false but we cannot have stub as false and stub_no_summary is true. 
If we want to use stub_no_summary in place of stub like you suggested here, then the NU has to change as well because I made stub as mandatory pre-requisite for stub_no_summary. Can you think this over once again and let me know?

README.md
saichint: ditto for the above comment

examples/cisco/demo_ospf.pp
saichint: same as above

tests/beaker_tests/cisco_ospf_area/test_ospf_area.rb
saichint: Yes. I did. The demo_manifest has both ipv4 and integer type area id.

tests/beaker_tests/cisco_ospf_area/test_ospf_area.rb
saichint: now that we are adding feature enable, we do not need this code

tests/beaker_tests/cisco_ospf_area/test_ospf_area.rb
saichint: No, similar to cisco_pim as we discussed today

README.md
chrisvanheuveln: IMO, by making it mandatory you place an unnecessary burden on the user. Allowing stub_no_summary to be used by itself provides flexibility. We have existing properties that support this model so I don't see the issue.

README.md
saichint: Added

lib/puppet/type/cisco_ospf_area.rb
saichint: ok

saichint:
Took care of stub/stub_no_summary behavior when one of both are given as discussed. Also added new test cases in beaker to reflect this.

PR #354, opened 2016-06-21
Add nssa properties to ospf_area provider

saichint:
This PR is for extending cisco_ospf_area with nssa parameters.
All tests pass for all nxos platforms.


README.md
```diff
@@ -2437,6 +2437,16 @@ cisco_ospf_area { 'my_ospf_instance my_vrf 1.1.1.1':
   filter_list_out => 'fout',
   stub            => true,
 }
+
+cisco_ospf_area { 'my_ospf_instance my_vrf 1000':
+  ensure                 => 'present',
+  nssa                   => 'md5',
```
chrisvanheuveln: `/md5/true/`

README.md
```diff
@@ -2463,6 +2473,25 @@ This is a route-map for filtering networks sent to this area. Valid values are s
 ##### `filter_list_out`
 This is a route-map for filtering networks sent from this area. Valid values are string or keyword 'default'.
 
+##### `nssa`
+This property defines the area as NSSA (not so stubby area). Valid values are true, false or keyword 'default'. This property is mutually exclusive with `stub` and `stub_no_summary`.
+
+##### `nssa_default_originate`
+This property configures to originate Type-7 default LSA into NSSA area. Valid values are true, false or keyword 'default'.
```
chrisvanheuveln: `Generates an NSSA External (type 7) LSA for use as a default route to the external autonomous system.`

README.md
```diff
@@ -2463,6 +2473,25 @@ This is a route-map for filtering networks sent to this area. Valid values are s
...
 
+##### `nssa`
+This property defines the area as NSSA (not so stubby area). Valid values are true, false or keyword 'default'. This property is mutually exclusive with `stub` and `stub_no_summary`.
+
+##### `nssa_default_originate`
+This property configures to originate Type-7 default LSA into NSSA area. Valid values are true, false or keyword 'default'.
+
+
+##### `nssa_no_redistribution`
+This property configures not to send redistributed LSAs into NSSA area. Valid values are true, false or keyword 'default'.
```
```
-This property configures not to send redistributed LSAs into NSSA area.
+Disable redistribution within the NSSA.
```

README.md
```diff
@@ -2463,6 +2473,25 @@ This is a route-map for filtering networks sent to this area. Valid values are s
...
+
+##### `nssa_no_summary`
+This property configures not to send summary LSAs into NSSA area. Valid values are true, false or keyword 'default'.
```
```
-This property configures not to send summary LSAs into NSSA area.
+Disables summary LSA flooding within the NSSA.
```

README.md
```diff
@@ -2463,6 +2473,25 @@ This is a route-map for filtering networks sent to this area. Valid values are s
...
+
+##### `nssa_route_map`
+This is the name of the Policy to control distribution of default route. Valid values are string or keyword 'default'. This MUST be set to 'default' when `nssa_default_originate` is 'default'.
```
```
-This is the name of the Policy to control distribution of default route.
+Filters external routes in the NSSA. This is the name of the route-map / route-policy. 
```

README.md
```diff
@@ -2463,6 +2473,25 @@ This is a route-map for filtering networks sent to this area. Valid values are s
...
+
+##### `nssa_translate_type7`
+This property translates LSA from type 7 to type 5. Valid values are 'always', 'always_supress_fa', 'supress_fa', 'never' or keyword 'default'.
```
chrisvanheuveln: This needs a little more info:

```
Translates NSSA external (type 7) LSAs to standard external (type 5) LSAs for use outside the NSSA. Valid values are one of the following keyword strings:
```

Keyword | Description
:--|:--
always             | Always translate
suppress_fa       | Forwarding Address Suppression
always_suppress_fa | Always translate & use Forwarding Address Suppression
never               | Never translate
default             | Translation is not configured

examples/cisco/demo_ospf.pp
```diff
@@ -86,4 +86,19 @@
...
+    filter_list_in         => 'bbb',
+    filter_list_out        => 'ccc',
+    nssa                   => true,
+    nssa_default_originate => true,
+    nssa_no_redistribution => true,
+    nssa_no_summary        => true,
+    nssa_route_map         => 'aaa',
+    nssa_translate_type7   => 'supress_fa',
+    range                  => [['10.3.0.0/16', '234'], ['10.3.1.1/32', 'not_advertise']],
+  }
```
chrisvanheuveln: IMO, we should keep these examples simple. You've already got examples above that show how to use authentication, cost, filter_list, range, etc so it just clutters up the nssa example by repeating them. I would just do this:

```
cisco_ospf_area { 'dark_blue my_nssa_vrf 5000':
    ensure                 => 'present',
    nssa                   => true,
    nssa_default_originate => true,
    nssa_no_redistribution => true,
    nssa_no_summary        => true,
    nssa_route_map         => 'aaa',
    nssa_translate_type7   => 'supress_fa',
}
```

lib/puppet/provider/cisco_ospf_area/cisco.rb
```diff
@@ -36,16 +43,22 @@
     :default_cost,
     :filter_list_in,
     :filter_list_out,
+    :nssa_route_map,
+    :nssa_translate_type7,
   ]
   OSPF_AREA_BOOL_PROPS = [
+    :nssa,
```
chrisvanheuveln: I assume you reordered `OSPF_AREA_ALL_PROPS` below because you want `:nssa` to process before the other nssa properties? If so then I would add a comment; e.g.

```
  OSPF_AREA_BOOL_PROPS = [
    :nssa, # :nssa should process before other nssa properties
    :nssa_default_originate,
```

lib/puppet/provider/cisco_ospf_area/cisco.rb
```diff
@@ -138,6 +151,48 @@ def properties_set(new_area=false)
...
+    vars = [
+      :nssa,
+      :nssa_route_map,
+      :nssa_default_originate,
+      :nssa_no_redistribution,
+      :nssa_no_summary,
+    ]
+    return unless vars.any? { |p| @property_flush.key?(p) }
+    # At least one var has changed, get all vals from manifest
+    attrs = {}
```
chrisvanheuveln: Duplicate of line 159

lib/puppet/provider/cisco_ospf_area/cisco.rb
```diff
@@ -138,6 +151,48 @@ def properties_set(new_area=false)
...
+    vars.each do |p|
+      if @resource[p] == :default
+        attrs[p] = @nu.send("default_#{p}")
+      else
+        attrs[p] = @resource[p]
+        attrs[p] = PuppetX::Cisco::Utils.bool_sym_to_s(attrs[p])
```
chrisvanheuveln: What's the reason for this utility call? Can't you just do this:

```
        attrs[p] = (@resource[p] == :true)
```

lib/puppet/provider/cisco_ospf_area/cisco.rb
```diff
@@ -138,6 +151,48 @@ def properties_set(new_area=false)
...
+    else
+      @nu.nssa_set(hash)
+      return
+    end
+    if attrs[:nssa_route_map] == '' || attrs[:nssa_route_map].nil?
+      hash[:route_map] = ''
+      hash[:rm] = ''
+    else
+      hash[:route_map] = 'route-map'
+      hash[:rm] = attrs[:nssa_route_map]
```
chrisvanheuveln: You shouldn't be setting actual cli syntax in the puppet code. This should be done in NU. I have some related comments for this in the following comment...

lib/puppet/provider/cisco_ospf_area/cisco.rb
```diff
@@ -138,6 +151,48 @@ def properties_set(new_area=false)
...
+    end
+    hash[:default_information_originate] = attrs[:nssa_default_originate] ? 'default-information-originate' : ''
+    hash[:no_redistribution] = attrs[:nssa_no_redistribution] ? 'no-redistribution' : ''
+    hash[:no_summary] = attrs[:nssa_no_summary] ? 'no-summary' : ''
+    @nu.nssa_set(hash)
   end
```
chrisvanheuveln: This entire method looks way more complicated than it needs to be. We should probably have a phone call to discuss. 

I would expect to see something like this instead:

```
  def nssa_set
    attrs = {}
    vars = [
      :nssa,
      :nssa_route_map,
      :nssa_default_originate,
      :nssa_no_redistribution,
      :nssa_no_summary,
    ]
    if vars.any? { |p| @property_flush.key?(p) }
      # At least one var has changed, get all vals from manifest
      vars.each do |p|
        val = @resource[p]
        val = @nu.send("default_#{p}") if val == :default
        val = val.to_s if val.is_a?(Symbol)
        attrs[p] = val
      end
    end
    return if attrs.empty?
    @nu.nssa_set(attrs)
  end
```

On the node utils side you'll need to update your yaml (remove the <rm> tag) and then fix route-map in nssa_set, i.e.:

```
+     hash[:route_map] = "route-map #{hash[:route_map]}" unless
+        hash[:route_map].to_s.empty?
       hash.keys.each do |k|
         @set_args[k] = hash[k]
       end
       config_set('ospf_area', 'nssa_set', @set_args)

```

lib/puppet/type/cisco_ospf_area.rb
```diff
@@ -27,16 +27,28 @@
...
+    cisco_ospf_area {'myrouter vrf2 2002':
+      ensure                  => 'present',
+      nssa                    => true,
+      nssa_default_originate  => true,
+      nssa_no_redistribution  => true,
+      nssa_no_summary         => true,
+      nssa_route_map          => 'rmap',
+      nssa_translate_type7    => 'always_supress_fa',
+      range                   => [['10.3.0.0/16', true, '23'],
+                                  ['10.3.3.0/24', false, '450']],
```
chrisvanheuveln: As with the demo manifest, just keep these simple.

lib/puppet/type/cisco_ospf_area.rb
```diff
@@ -124,6 +136,44 @@ def name
     munge { |value| value == 'default' ? :default : value }
   end # property filter_list_out
 
+  newproperty(:nssa) do
+    desc 'Defines the area as NSSA (not so stubby area). This is
+          mutually exclusive with stub and stub_no_summary.'
```
chrisvanheuveln: Please update these `desc` entries to match the readme comments I made.

saichint: done with fixes for all comments. Also created another PR for NU for the nssa_set method changes : https://github.com/cisco/cisco-network-node-utils/pull/434

PR #355, opened 2016-06-29
router_ospf_area_vlink: new provider

saichint:
This PR is for cisco_ospf_area_vlink provider. 
All demo_manifests and beaker tests passed on all nexus platforms.


lib/puppet/type/cisco_ospf_area_vlink.rb
```diff
@@ -0,0 +1,248 @@
...
+          'cleartext', '3des' or 'cisco_type_7' encryption,
+          and 'default', which defaults to 'cleartext'."
+
+    newvalues(:cleartext,
+              :'3des',
+              :cisco_type_7,
+              :default)
+
+    munge do |value|
+      value = :cleartext if value.to_sym == :default
```
chrisvanheuveln: This seems odd. Why is it necessary to convert default? Why not just do:  ` munge(&:to_sym)`  ?

lib/puppet/type/cisco_ospf_area_vlink.rb
```diff
@@ -0,0 +1,248 @@
...
+
+    munge { |value| value == 'default' ? :default : value }
+  end # property authentication_key_password
+
+  newproperty(:dead_interval) do
+    desc "Sets the time in seconds that a neighbor waits for a Hello packet
+          before declaring the local router as dead and tearing down
+          adjacencies. Valid values are integer, keyword 'default'."
+
+    munge { |value| value == 'default' ? :default : Integer(value) }
```
chrisvanheuveln: Curious about the `Integer(value)`. Most of the time our NU infra is returning string values so I would think that you should do `value.to_s` instead.

lib/puppet/type/cisco_ospf_area_vlink.rb
```diff
@@ -0,0 +1,248 @@
...
+          'cleartext', '3des' or 'cisco_type_7' encryption,
+          and 'default', which defaults to 'cleartext'."
+
+    newvalues(:cleartext,
+              :'3des',
+              :cisco_type_7,
+              :default)
+
+    munge do |value|
+      value = :cleartext if value.to_sym == :default
```
chrisvanheuveln: ditto why convert default?

tests/beaker_tests/cisco_ospf_area_vlink/test_ospf_area_vlink.rb
```diff
@@ -0,0 +1,111 @@
...
+  },
+}
+
+#################################################################
+# TEST CASE EXECUTION
+#################################################################
+test_name "TestCase :: #{tests[:resource_name]}" do
+  # -------------------------------------------------------------------
+  device = platform
+  logger.info("#### This device is of type: #{device} #####")
```
chrisvanheuveln: This logger msg is unnecessary. The platform call logs the device type already.

chrisvanheuveln:
Just a few minor comments. Code looks good overall.

tests/beaker_tests/cisco_ospf_area_vlink/test_ospf_area_vlink.rb
```diff
@@ -0,0 +1,111 @@
...
+  logger.info("\n#{'-' * 60}\nSection 1. Default Property Testing")
+  test_harness_run(tests, :default)
+
+  id = :default
+  tests[id][:ensure] = :absent
+  tests[id].delete(:preclean)
+  test_harness_run(tests, id)
+
+  # -------------------------------------------------------------------
+  logger.info("\n#{'-' * 60}\nSection 2. Non Default Property Testing")
```
mikewiebe: We need title pattern testing for this provider.  Quite a few possibilities here.

saichint: Since all namevar in the tile pattern are mandatory, leaving this as it is,

lib/puppet/type/cisco_ospf_area_vlink.rb
saichint: The NU is returning int type and so need to convert this to Integer if a string is given in the manifest.

saichint:
Changed the code for review comments where necessary and retested.

tests/beaker_tests/cisco_ospf_area_vlink/test_ospf_area_vlink.rb
```diff
@@ -0,0 +1,110 @@
...
+    transmit_delay:                     '400',
+  },
+}
+
+#################################################################
+# TEST CASE EXECUTION
+#################################################################
+test_name "TestCase :: #{tests[:resource_name]}" do
+  # -------------------------------------------------------------------
+  platform
```
chrisvanheuveln: Not needed. `platform` gets called during `test_harness_run`

PR #357, opened 2016-07-06
Fixes for cisco_interface_ospf provider

saichint:
This PR is for fixing cisco_interface_ospf provider/type.
1. All parameters now accept 'default' keyword which was missing for many of them before.
2. message_digest_encryption_type and message_digest_algorithm_type can be changed by themselves without changing the message_digest_key_id or message_digest_password as they are now included in the puppet resource.
3. cleaned the the provider/type code to follow latest code guidelines.
4. Removed the auto-requires code from type
5. Removed all the older beaker tests as they are not properly testing message-digest parameters.
6. Added new beaker test which tests all parameters and follows the latest guidelines.

All beaker tests and demo manifests pass across all nxos platforms.


tests/beaker_tests/cisco_interface_ospf/test_interface_ospf.rb
```diff
@@ -0,0 +1,126 @@
...
+  code:           [0, 2],
+  resource:       {
+    area:                           '0.0.0.200',
+    cost:                           0,
+    dead_interval:                  40,
+    hello_interval:                 10,
+    message_digest:                 'false',
+    message_digest_key_id:          0,
+    message_digest_algorithm_type:  'md5',
+    message_digest_encryption_type: 'cleartext',
```
robert-w-gries: Is `message_digest_password` missing a `resource` entry?

saichint: No. the default for message_digest_password is nil. So no need to specify that.
Check this:
https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/cmd_ref/interface_ospf.yaml#L54

robert-w-gries:
👍

tests/beaker_tests/cisco_interface_ospf/test_interface_ospf.rb
```diff
@@ -0,0 +1,126 @@
...
+    message_digest_password:        '046E1803362E595C260E0B240619050A2D',
+    passive_interface:              'true',
+  },
+  resource:       {
+    area: '0.0.0.200'
+  },
+}
+
+# Overridden to properly handle dependencies for this test file.
+def test_harness_dependencies(_tests, _id)
```
chrisvanheuveln: It looks like this really only needs to process at the beginning of the test. If so just add an early return:

`  return unless id == :default`

lib/puppet/provider/cisco_interface_ospf/cisco.rb
```diff
@@ -31,29 +31,28 @@
...
-    :hello_interval, :dead_interval
+  INTF_OSPF_NON_BOOL_PROPS = [
+    :cost,
+    :hello_interval,
+    :dead_interval,
+    :message_digest_key_id,
+    :message_digest_algorithm_type,
+    :message_digest_encryption_type,
+    :message_digest_password,
+    :area,
```
chrisvanheuveln: nit: we like to keep the properties in alpha order, except when they need to process in a certain order -- in which case we call out why it's misordered. Does `area` need special handling?

saichint: I am not sure why it was done this way. I retried after reordering, and it seems ok. On the node, the CLI also seems ok without any order issues. May be 'area' property was added later and it was appended in the end? Anyway, took care of it.

PR #381, opened 2016-08-17
beaker fix for resilient and symmetry non support on n3k

saichint:
On some n3k flavors, symmetry and resilient properties are not supported. This PR is to take care of the beaker tests. All tests passed on relevant platforms.


tests/beaker_tests/cisco_portchannel_global/test_portchannel_global.rb
```diff
@@ -273,6 +273,42 @@
   },
 }
 
+tests['default_properties_no_resilient'] = {
+  title_pattern:  'default',
+  manifest_props: "
+    bundle_hash                  => 'default',
+    bundle_select                => 'default',
+    symmetry                     => 'default',
+  ",
```
chrisvanheuveln: This testfile is getting seriously unwieldy. :-(

The only difference between test cases `default_properties_no_rotate` and `default_properties_no_resilient` is that one has a resilient property (based on hardware support) and the boolean value is different for symmetry. This could be simplified a lot if you convert the testfile to use `unsupported_properties` and `test_harness_run`; in which case you'll have one testcase for `_no_rotate` that automatically removes the unsupported properties.

tests/beaker_tests/cisco_portchannel_global/test_portchannel_global.rb
```diff
@@ -469,49 +509,91 @@ def test_harness_portchannel_global(tests, id)
...
+
+      tests[id][:desc] = '2.4 Non Default Properties'
+      tests['non_default_properties_no_resilient'][:manifest_props]['mac'] = 'port'
+      tests['non_default_properties_no_resilient'][:resource_props]['bundle_hash'] = 'port'
+      test_harness_portchannel_global(tests, id)
+
+      tests[id][:desc] = '2.5 Non Default Properties'
+      tests['non_default_properties_no_resilient'][:manifest_props]['port'] = 'port-only'
+      tests['non_default_properties_no_resilient'][:resource_props]['bundle_hash'] = 'port-only'
+      tests['non_default_properties_no_resilient'][:manifest_props]['dst'] = 'src-dst'
```
chrisvanheuveln: Most if not all of this goes away with a rewrite, but it caught my eye. You don't have `port`, `src`, `dst`, `mac`, etc property names in your type, so what are these settings doing?

saichint:
Rewrote the test file completely using the latest template. Tested on all nxos platforms, including different flavors of n3k. All tests pass

tests/beaker_tests/cisco_portchannel_global/test_portchannel_global.rb
```diff
@@ -12,509 +12,323 @@
...
-  }
-EOF"
+tests[:non_default][:manifest_props].merge!(manifest_non[:n3k]) if platform[/n3k/]
+tests[:non_default][:manifest_props].merge!(manifest_non[:n56k]) if platform[/n(5|6)k/]
+tests[:non_default][:manifest_props].merge!(manifest_non[:n8k]) if platform[/n8k/]
+tests[:non_default][:manifest_props].merge!(manifest_non[:n9k]) if platform[/n9k/]
+
+def unsupported_properties(tests, _id)
+  unprops = []
+  if platform[/n7k/]
```
chrisvanheuveln: nit: I prefer a case here instead of all these elsif's. I won't make you change it though :-)

tests/beaker_tests/cisco_portchannel_global/test_portchannel_global.rb
```diff
@@ -12,509 +12,323 @@
...
+      :symmetry
+  elsif platform[/n9k/]
+    unprops <<
+      :asymmetric <<
+      :hash_distribution <<
+      :hash_poly <<
+      :load_defer
+  end
+  unprops << :resilient << :symmetry if
+    platform[/n3k/] && tests[:resilient_unsupported]
```
chrisvanheuveln: The n3k check is probably unnecessary.

tests/beaker_tests/cisco_portchannel_global/test_portchannel_global.rb
```diff
@@ -12,509 +12,323 @@
...
+    resource_probe(agent,
+                   'cisco_portchannel_global default resilient=true',
+                   'ERROR: This feature is not supported on this platform.')
 end
 
 #################################################################
 # TEST CASE EXECUTION
 #################################################################
-test_name "TestCase :: #{testheader}" do
+test_name "TestCase :: #{tests[:resource_name]}" do
```
chrisvanheuveln: You should have a teardown and cleanup at the beginning here, e.g. https://github.com/cisco/cisco-network-puppet-module/blob/develop/tests/beaker_tests/cisco_vlan/test_vlan.rb#L127

This is a new format we recently added. The teardown will run even if the test hits an error. The initial cleanup is optional: you can either do it here or rely on a `:preclean` entry in the testcase. Either way you should have some kind of cleanup before & after this test runs.

chrisvanheuveln: Just a few minor things overall. This rewrite is a huge improvement over the old code!

tests/beaker_tests/cisco_portchannel_global/test_portchannel_global.rb
saichint: Ok. left it as it is.

tests/beaker_tests/cisco_portchannel_global/test_portchannel_global.rb
saichint: Added teardown where it calls the provider default test case.

PR #383, opened 2016-08-25
Rework dns test to save/restore configs

chrisvanheuveln:
* We weren't running the existing test because it messes up live DNS settings on the testbeds
* This commit saves any original DNS settings and restores them at teardown time.
* This is a NXOS-only fix. It's a little clunky but I did it this way to avoid affecting the XR test.

### DIFF
--- a/tests/beaker_tests/netdev_stdlib/network_dns_provider_nondefaults.rb
+++ b/tests/beaker_tests/netdev_stdlib/network_dns_provider_nondefaults.rb
@@ -77,46 +77,51 @@ def check_dns_warning
   fail dns_warning if dns_warning
 end
 
-# Helper for XR testbed cleanup
-def xr_dns_clean(agent)
-  return unless operating_system == 'ios_xr'
-  # remove any existing resources that we will be testing against
-  resource_titles(agent, :domain_name, :clean)
-  ['no domain list test.com',
-   'no domain list test.net',
-   'no domain name switch1.test.com',
-   'no domain name switch2.test.com',
-   'no domain name-server 2001:4860:4860::8888',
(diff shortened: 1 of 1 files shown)


tests/beaker_tests/netdev_stdlib/network_dns_provider_nondefaults.rb
```diff
@@ -58,64 +58,79 @@
...
+                  :array)
+  return nil unless cmds
+
+  no_cmds = cmds.map { |cmd| "no #{cmd}" }.join(' ; ')
+  test_set(agent, no_cmds)
+  # return original commands as str
+  cmds.join(' ; ')
+end
+
+def dns_save(agent)
```
mikewiebe: So the changes look good, but did you test to make sure that entries that bleed through into the /etc/resolve.conf file really get removed?  When I worked on the minitest side of this, I found that I had to backup/restore the resolve.conf file to make sure none of the bleed through entries remained.

chrisvanheuveln: No I didn't check that. I thought the original problem was with the vsh commands being removed. I'll check out the bleed-thru as well.

chrisvanheuveln:
I ran into an image bug where `no ip domain-name foo` does not always update /etc/resolv.conf. As a result my code does the save, cleanup, restore just fine and the test passes just fine, but if you go back to the cli later and add a dns command then boom! there are suddenly a bunch of stale entries from the beaker test now residing in /etc/resolv.conf. So I'm going to close this for now and try to get the image defect corrected.

PR #389, opened 2016-09-06
cisco_interface enhancement

saichint:
This PR is for enhancing the cisco_interface provider by adding dhcp relay parameters and storm control parameters. 
All tests pass on all nexus platforms


README.md
```diff
@@ -1974,6 +1978,15 @@ and 'default'.
 Enable/Disable negotiate auto on the interface. Valid values are 'true',
 'false', and 'default'.
 
+##### `storm_control_broadcast`
+Allowed broadcast traffic level. Valid values are string, keyword 'default'.
```
mikewiebe: I think it would be helpful to indicate what the string represents.  `Valid values are a string representing the broadcast level or keyword 'default'.`

examples/cisco/demo_interface.pp
```diff
@@ -36,6 +36,16 @@
       ensure => 'present',
     }
 
+    $ipv4_dhcp_relay_info_trust = platform_get() ? {
+      /(n3k|n7k|n8k|n9k)/ => true,
```
mikewiebe: Remove the `n8k` reference.  We will likely need to scrub these before the release.

lib/puppet/type/cisco_interface.rb
```diff
@@ -1042,4 +1053,144 @@ def insync?(is)
...
+  end # property ipv4_dhcp_relay_addr
+
+  newproperty(:ipv4_dhcp_relay_info_trust) do
+    desc 'Enables relay trust on this interface.'
+
+    newvalues(:true, :false, :default)
+  end # property ipv4_dhcp_relay_info_trust
+
+  newproperty(:ipv4_dhcp_relay_src_addr_hsrp) do
+    desc 'Enables VIP instead of SVI address'
```
mikewiebe: Use `Virtual IP` instead of `VIP`

lib/puppet/type/cisco_interface.rb
```diff
@@ -1042,4 +1053,144 @@ def insync?(is)
...
+      value.inspect
+    end
+
+    def is_to_s(value)
+      value.inspect
+    end
+
+    munge do |value|
+      begin
+        return value = :default if value == 'default'
```
mikewiebe: Simplify: `return :default if value == 'default'`

examples/cisco/demo_interface.pp
saichint: Ok, leaving it as it

PR #395, opened 2016-09-28
event_history_periodic changes for bgp

saichint:
This PR is for fixing event_history_periodic property of bgp as the defaults is changed in I5 image. The new default is 'false' whereas the earlier images it is 'size_small'. The changes are backward compatible with older releases. Also all event_history properties will now accept size in bytes. The CLIs were accepting th size in bytes in the older releases as well using hidden commands but now they are not hidden, so the changes are transparent for all releases.


README.md
```diff
@@ -1038,6 +1038,7 @@ Manages configuration of a BGP instance.
 |:--------|:-------------|
 | `disable_policy_batching_ipv4` | Not supported on N5k, N6k, N7k |
 | `disable_policy_batching_ipv6` | Not supported on N5k, N6k, N7k |
+| `event_history_periodic      ` | default is 'false' on images I5 and above |
 | `neighbor_down_fib_accelerate` | Not supported on N5k, N6k, N7k |
 | `reconnect_interval`           | Not supported on N5k, N6k, N7k |
 | `suppress_fib_pending`         | Idempotence supported only on I5 images |
```
mikewiebe: Let's change this to be more explicit: `Idempotence supported only on 7.0(3)I5(1) and later images on n3k and n9k`

lib/puppet/provider/cisco_bgp/cisco.rb
```diff
@@ -209,6 +209,28 @@ def timer_bestpath_limit_set
     @bgp_vrf.timer_bestpath_limit_set(limit, always)
   end
 
+  def image_older_version
```
mikewiebe: I would prefer the method name `legacy_image?`

lib/puppet/provider/cisco_bgp/cisco.rb
```diff
@@ -209,6 +209,28 @@ def timer_bestpath_limit_set
...
+    fd = Facter.value('cisco')
+    image = fd['images']['system_image']
+    pid = fd['inventory']['chassis']['pid']
+    image[/7.0.3.I2|I3|I4/] || pid[/N(5|6|7|8)/]
```
mikewiebe: Is there a reason we are not using the image version for N5|6|7|8k platforms?  If possible let's use the image version so that if/when images on these platforms support the new cli we don't have to make a change in the code base to enable the new cli.

tests/beaker_tests/cisco_bgp/test_bgp.rb
```diff
@@ -103,6 +103,16 @@
   },
 }
 
+# older_version default value
+resource = {
+  old: {
```
mikewiebe: Let's change this to `legacy:`

README.md
```diff
@@ -1038,6 +1038,7 @@ Manages configuration of a BGP instance.
 |:--------|:-------------|
 | `disable_policy_batching_ipv4` | Not supported on N5k, N6k, N7k |
 | `disable_policy_batching_ipv6` | Not supported on N5k, N6k, N7k |
+| `event_history_periodic      ` | default is 'false' on images I5 and above |
```
saichint: done

README.md
saichint: Done

lib/puppet/provider/cisco_bgp/cisco.rb
saichint: Done

lib/puppet/provider/cisco_bgp/cisco.rb
saichint: not doing it as we discussed

tests/beaker_tests/cisco_bgp/test_bgp.rb
saichint: Done

PR #400, opened 2016-10-07
bgp event_history

saichint:
This PR is for fixing few issues with event history properties as their defaults and behaviors have changed in the latest platform code release. Also added two new properties to event_history (errors and objstore) to the bgp provider which are available in the latest platform release.

### DIFF
--- a/examples/cisco/demo_bgp.pp
+++ b/examples/cisco/demo_bgp.pp
@@ -58,6 +58,22 @@ class ciscopuppet::cisco::demo_bgp {
     default => undef
   }
 
+  $event_history_errors = platform_get() ? {
+    /(n3k|n9k)/ => $facts['cisco']['images']['system_image'] ? {
+      /(I2|I3|I4)/ => undef,
+      default => 'size_small'
+    },
+    default => undef
+  }
+
+  $event_history_objstore = platform_get() ? {
+    /(n3k|n9k)/ => $facts['cisco']['images']['system_image'] ? {
+      /(I2|I3|I4)/ => undef,
+      default => 'size_small'
+    },
+    default => undef
+  }
+
   cisco_bgp { '55.77 default':
     ensure                                 => present,
 
@@ -73,11 +89,9 @@ class ciscopuppet::cisco::demo_bgp {
     enforce_first_as                       => false,
     event_history_cli                      => 'size_small',
     event_history_detail                   => 'size_medium',
-    # event_history_errors is not supported on all platforms
-    # event_history_errors                   => 'size_small',
+    event_history_errors                   => $event_history_errors,
     event_history_events                   => 'size_large',
-    # event_history_objstore is not supported on all platforms
-    # event_history_objstore                 => 'size_large',
+    event_history_objstore                 => $event_history_objstore,
     event_history_periodic                 => 'size_disable',
     maxas_limit                            => $maxas_limit,
     suppress_fib_pending                   => $suppress_fib_pending,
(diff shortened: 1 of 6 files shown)


examples/cisco/demo_bgp.pp
```diff
@@ -73,7 +73,11 @@
     enforce_first_as                       => false,
     event_history_cli                      => 'size_small',
     event_history_detail                   => 'size_medium',
+    # event_history_errors is not supported on all platforms
+    # event_history_errors                   => 'size_small',
```
mikewiebe: Why not add a conditional check and set to `undef` on unsupported platforms?

lib/puppet/provider/cisco_bgp/cisco.rb
```diff
@@ -216,17 +218,68 @@ def legacy_image?
...
+    @property_hash[:event_history_cli]
+  end
+
+  def event_history_cli=(should_value)
+    should_value = @bgp_vrf.default_event_history_cli if should_value == 'default' || should_value == 'true'
+    should_value = should_value.to_sym unless should_value =~ /\A\d+\z/
+    @property_flush[:event_history_cli] = should_value
+  end
+
+  def event_history_detail
```
mikewiebe: IMO, this code would be a bit easier to read and understand as follows:


```
# Compare @property_hash to default getter helper method
def default?(prop)
  return @property_hash[prop.to_sym] == @bgp_vrf.send("default_#{prop}")
end

def event_history_detail
  case resource[:event_history_detail]
    when 'default'
      return 'default' if default?(:event_history_detail) 
    when 'size_disable'
      return 'size_disable' if default?(:event_history_detail) && !legacy_image?
    else
      return @property_hash[:event_history_detail]
  end
end

**Disclaimer** I have not tested this code :)

Comment (mikewiebe) [inline lib/puppet/provider/cisco_bgp/cisco.rb]: If this works then use it to simplify other custom getters in this provider.

File: examples/cisco/demo_bgp.pp (hunk 1 above)
Comment (saichint) [inline examples/cisco/demo_bgp.pp]: is there a version check condition example for demo manifests? I could not find it, can you please point me to one?

Comment (mikewiebe) [inline examples/cisco/demo_bgp.pp]: Good point.  There isn't one but I think it would be good to add one under https://github.com/cisco/cisco-network-puppet-module/tree/develop/lib/puppet/parser/functions.

Just add a new function say `image_version_get.rb`.

You can use `platform_get.rb` as an example.  When you add the new function, you will need to restart your puppet master to get it to detect the new function.

File: lib/puppet/provider/cisco_bgp/cisco.rb (hunk 2 above)
Comment (saichint) [inline lib/puppet/provider/cisco_bgp/cisco.rb]: made some minor changes to your code but the design is same. Works now

File: examples/cisco/demo_bgp.pp (hunk 1 above)
Comment (saichint) [inline examples/cisco/demo_bgp.pp]: I made the changes to get 2 conditions using facts and platform_get(), this is working! But I don't know if there is a better way, I tried other combinations but I am getting syntax failures. If there are more than 2 conditions, we can keep creating loops inside but not sure if that looks good but at least there seems to be a way. If you have a better suggestion, let me know, I can try.

File: examples/cisco/demo_bgp.pp [hunk 3]
```diff
@@ -58,6 +58,22 @@
     default => undef
   }
 
+  $event_history_errors = platform_get() ? {
```
mikewiebe: Nice work :)

PR #401, opened 2016-10-18
Rel141/lacp suspend

mikewiebe:
**Summary of changes:**
* Update beaker test to verify the correct default value on `n3k` and other nxos platforms.
* Update caveat for lacp_suspend property to indicate the the interface is automatically shutdown on `n9k` platforms.

Tested on: n3k(I2, I5), n9k(I5)

### DIFF
--- a/tests/beaker_tests/cisco_interface_portchannel/test_interface_portchannel.rb
+++ b/tests/beaker_tests/cisco_interface_portchannel/test_interface_portchannel.rb
@@ -105,7 +105,7 @@ tests[:non_default_sym] = {
     lacp_graceful_convergence: 'false',
     lacp_max_bundle:           '10',
     lacp_min_links:            '3',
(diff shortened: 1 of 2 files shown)


tests/beaker_tests/cisco_interface_portchannel/test_interface_portchannel.rb
```diff
@@ -90,7 +90,7 @@
     'lacp_graceful_convergence' => 'true',
     'lacp_max_bundle'           => '32',
     'lacp_min_links'            => '1',
-    'lacp_suspend_individual'   => 'true',
+    'lacp_suspend_individual'   => platform[/n3k/] ? 'false' : 'true',
```
saichint: please do similar change for non default case as well

saichint:
:+1:

PR #407, opened 2016-11-17
Rel150/product tag

mikewiebe:
Adds a helper method `product_tag` in cmnutls.rb that returns a short but meaningful product tag for nxos platforms.


saichint:
:+1:

PR #409, opened 2016-12-01
Rel150/atherton fretta fixes

mikewiebe:
**Summary of changes:**
- Add new utility method `add_quotes` that adds starting/trailing double quotes to a string.
  - Used to match strings in cases where nxapi preserves the double quotes.
- Modified beaker tests that call the `add_quotes` method for the following reason.
  - Calling the `add_quotes` method updates the stdout buffer with the image version and overwrites the output of the `puppet resource` command used for the test.
  - Update saves the stdout buffer to be used in the test.
- Updates utility method `nexus_image` to handle Atherton versions and be more generally flexible.

**Tested on:** n7k(helsinki and atherton), n9k evergreen


saichint:
:+1:

PR #410, opened 2016-12-01
pim and interface bfd

saichint:
This PR is for adding bfd property to cisco_pim and pim_bfd property to cisco_interface providers. All tests pass.


mikewiebe:
👍

PR #411, opened 2016-12-08
n7k atherton hsrp

saichint:
This PR is for adding hsrp support for n7k running atherton image. The README.md file is NOT changed yet, as there are no references to atherton in that. But it needs to be specified that this hsrp support is only for n7k running atherton. Please provide any comment on how to show that in the caveats.

### DIFF
--- a/examples/cisco/demo_hsrp.pp
+++ b/examples/cisco/demo_hsrp.pp
@@ -26,7 +26,16 @@ class ciscopuppet::cisco::demo_hsrp {
     extended_hold => 200,
   }
 
-  if platform_get() =~ /(n3k|n9k)/ {
+  $n7k_ath = platform_get() ? {
+    'n7k' => $facts['cisco']['images']['system_image'] ? {
+      /(8.0)/ => true,
+      default => false
+    },
+    /(n3k|n9k)/ => true,
+    default => false
+  }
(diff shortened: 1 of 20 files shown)


mikewiebe:
@saichint Given that we don't know exactly what the Atherton version number will be, just add the following in the caveats table and we will update it with the exact version string later.

`Supported in OS Version 8.0 and later on N7k`

examples/cisco/demo_hsrp.pp
```diff
@@ -47,6 +56,12 @@
       hsrp_version       => 2,
     }
 
+    cisco_command_config { 'ipv6-addr':
```
mikewiebe: Why do we need to use cisco_command_config here to add the ipv6 address?  Can't we use cisco_interface for this?

saichint: I could not find any property with which I can set ipv6 address in the cisco_interface provider. I do see ipv4 address but no ipv6. If I am missing something, please correct me and point me to the right property.

mikewiebe: @saichint your right.  We don't have support for ipv6.  Never mind.

mikewiebe:
👍

Issue #415, opened 2017-01-24
Churn: HSRP Group resource type


abrader:
Hello friends at Cisco!

Got another for ya...this one is similar to [SNMP Community - Symbols Cause Churn](https://github.com/cisco/cisco-network-puppet-module/issues/414)

When customer calls HSRP Group creation via your resource type:

```puppet
cisco_interface_hsrp_group { $hsrp_title :
  ensure      => present,
  ipv4_enable => true,
  ipv4_vip    => $vlans[$vlan_id]['hsrp_vip'],
}
```

We see these resultant lines in the log even after repeated follow-up runs (where we have subsequently confirmed with show commands that the respective HSRP group(s) exist):

```
Notice: /Stage[main]/CustomerModule::Mg::Vlans/Cisco_interface_hsrp_group[VlanXXX Y ipv4]/ensure: created
```

This is likely due to the matching that is being performed here that is not catching the previously created HSRP Group(s) correctly:  https://github.com/cisco/cisco-network-puppet-module/blob/develop/lib/puppet/provider/cisco_interface_hsrp_group/cisco.rb#L138L140

saichint:
Hi Andrew,
Thanks for bringing this to attention.
The name of the interface_hsrp_group is currently case-sensitive. That is why you are seeing the problem.
If you change the name to have “vlan …” instead of “Vlan …” (lowercase v), then it will be fine.
We will make the change so that the name won’t be case sensitive in our next release.

Thanks
Sai

mikewiebe:
@saichint Can you include test results here with the fix?  Then we can close this issue.

saichint:
Manifest:
cisco_interface_hsrp_group { 'Vlan11 11 ipv4':
      ensure                   => 'present',
      ipv4_enable           => true,
      ipv4_vip                 => '2.2.2.2',
    }

Test results:

First puppet run:
Notice: /Stage[main]/Ciscopuppet::Cisco::Demo_hsrp/Cisco_interface_hsrp_group[Vlan11 11 ipv4]/ensure: created
Notice: Applied catalog in 9.01 seconds

Second run:
Notice: Applied catalog in 3.99 seconds

PR #416, opened 2017-01-24
fix the case sensitive issue for interface_hsrp_group

saichint:
Summary:
This update fixes the issue with the cisco_interface_hsrp_group provider described in cisco/cisco-network-puppet-module#415

As the name of the interface is case sensitive, it was causing idempotent issues. Fixed


mikewiebe:
👍

Issue #423, opened 2017-02-09
Feature request; "load-interval" for cisco_interface


tomcooperca:
Hi,
Can you please add the "load-interval" attribute to the cisco_interface type?
I need to manage these settings to keep interface timers consistent across all our switches. Thanks!

mikewiebe:
@tomcooperca We can certainly add that as a managed attribute.  We will schedule this work in an upcoming sprint and update the case when the work is complete.

tomcooperca:
@mikewiebe thank you!

mikewiebe:
@tomcooperca Code has been committed for this feature into our develop branch and will be part of the `1.6.0` version of the puppet module.

PR #425, opened 2017-02-15
Add load-interval props to cisco_interface

saichint:
This PR is for adding load-interval properties to cisco_interface provider. All tests pass on all platforms.
This update adds feature request; "load-interval" for cisco_interface #423


lib/puppet/type/cisco_interface.rb
```diff
@@ -1251,4 +1254,29 @@ def is_to_s(value)
 
     munge { |value| value == 'default' ? :default : Integer(value) }
   end # property hsrp_version
+
+  ############################
+  # load-interval attributes #
+  ############################
+
+  newproperty(:load_interval_counter_1_delay) do
+    desc "Load interval delay for counter 1 in sec. Valid values are
```
mikewiebe: *Nit:* Just specify `seconds` instead of `sec.`

PR #426, opened 2017-02-22
fix double quotes issue for passwords

saichint:
This PR is for fixing double quotes issue on tacacs and radius providers. The fix is mainly to override the provider get method and compare the password against the resource, and if the resource has no double quotes, then get rid of them in the get value. Also fixed the beaker tests to add the quotes properly in the resource area.


tests/beaker_tests/cisco_tacacs_server/tacacsserver_provider_nondefaults.rb
```diff
@@ -100,7 +100,7 @@
                              { 'ensure'              => 'present',
                                'timeout'             => '50',
                                'deadtime'            => '0',
-                               'encryption_password' => add_quotes('WXYZ12'),
```
mikewiebe: Go ahead and remove the `add_quotes` method assuming it's not used anywhere else.

lib/puppet/provider/cisco_tacacs_server/cisco.rb
```diff
@@ -114,6 +114,15 @@ def destroy
     @tacacs_server = nil
   end # destroy
 
+  def encryption_password
```
mikewiebe: BTW, for cisco providers we can take care of this in the type.  We only need the special getter code in the netdev providers.

saichint:
Done with changes

PR #427, opened 2017-02-27
Feature/upgrade services

rahushen:
Added new type and provider for cisco_service (to upgrade device)


mikewiebe:
@rahushen Looks like you have some rubocop errors.

lib/puppet/provider/cisco_service/cisco.rb
```diff
@@ -0,0 +1,128 @@
+#
+# Puppet provider for feature service
+#
+# Copyright (c) 2014-2016 Cisco and/or its affiliates.
```
mikewiebe: `Copyright (c) 2017...`

lib/puppet/type/cisco_service.rb
```diff
@@ -0,0 +1,117 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+Puppet::Type.newtype(:cisco_service) do
+  @doc = "Manages the version of Cisco Image running on a device.
+
+  ```
+  cisco_service {\"<version>\":
```
mikewiebe: While services is the larger bucket for router utilites supported by puppet, it would be better to call the type `cisco_upgrade` instead of `cisco_service`.

lib/puppet/provider/cisco_service/cisco.rb
```diff
@@ -0,0 +1,124 @@
+#
+# Puppet provider to manage upgrade of Cisco devices
+#
+# Copyright (c) 2016-2017 Cisco and/or its affiliates.
```
mikewiebe: Sorry.... my fault.. this should be just `2017`.

lib/puppet/type/cisco_upgrade.rb
```diff
@@ -0,0 +1,117 @@
...
+    desc 'Name of the version of image. Valid values are string.'
+  end # param version
+
+  newparam(:force_all) do
+    desc 'Force upgrade the device.'
+    defaultto :false
+    newvalues(:true, :false)
+  end # param force_all
+
+  newparam(:delete_boot) do
```
mikewiebe: I would be a bit more explicit with the name of the parameter: `delete_boot_image`

lib/puppet/type/cisco_upgrade.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  newparam(:name) do
+    desc 'Name of cisco_service, not used, but needed for puppet'
+  end
+
+  newparam(:version, namevar: :true) do
+    desc 'Name of the version of image. Valid values are string.'
+  end # param version
+
+  newparam(:force_all) do
```
mikewiebe: IMO, `:force_upgrade` or `:upgrade_force` is more clear.

lib/puppet/type/cisco_upgrade.rb
```diff
@@ -0,0 +1,117 @@
...
+    newvalues(:true, :false)
+  end # param delete_boot
+
+  ##############
+  # Attributes #
+  ##############
+
+  newproperty(:source_uri) do
+    desc 'URI to the image to install on the device. Format <media>:<image>.
+          Valid values are string.'
```
mikewiebe: I would add a calls to the `validate` and or `munge` methods to make sure the format of the version is correct.  No white space, unexpected control characters.. etc..

lib/puppet/type/cisco_upgrade.rb
```diff
@@ -0,0 +1,117 @@
...
+  end # property source_uri
+
+  def check_version
+    # validate that the version string is not empty or nil
+    # and that only the version string consists of only a
+    # few permitted characters
+    fail ArgumentError,
```
mikewiebe: Use `validate` and/or `munge` methods for these kinds of checks instead of `fail ArgumentError`.  `validate` is used for input validation and `munge` is used to make sure the data is in the correct format after being validated.

lib/puppet/type/cisco_upgrade.rb
```diff
@@ -0,0 +1,117 @@
...
+         'Invalid version string. Version can only have the following
+          characters: 0-9, a-z, A-Z, (, ) and .' unless
+         (/([0-9a-zA-Z().]*)/.match(self[:version]))[0] == self[:version]
+  end
+
+  def check_source_uri
+    # validate that the source_uri property is specified as
+    # <media>:<filename>
+    image = self[:source_uri].split(':')
+    fail ArgumentError,
```
mikewiebe: `validate and/or munge`

lib/puppet/type/cisco_upgrade.rb
```diff
@@ -0,0 +1,117 @@
...
+         'source_uri should be of the format <media>:<filename>' if
+      image[0] == image[-1]
+  end
+
+  # Validation block
+  validate do
+    fail ArgumentError,
+         'source_uri is required' if self[:source_uri].nil?
```
mikewiebe: `validate and/or munge`

lib/puppet/type/cisco_upgrade.rb
```diff
@@ -0,0 +1,123 @@
...
+  newparam(:name, namevar: :true) do
+    # Parameter used only to satisfy namevar
+    desc 'Name of cisco_upgrade instance. Valid values are string'
+    validate do |name|
+      warning "only 'image' is accepted as a valid name" if name != 'image'
+    end
+  end
+
+  newparam(:source_uri) do
+    desc 'URI to the image to install on the device. Format <media>:<image>.
```
mikewiebe: *nit:* Simplify to `Source image URI.  Format <uri:image>`

lib/puppet/type/cisco_upgrade.rb
```diff
@@ -0,0 +1,123 @@
...
+    validate do |uri|
+      fail "source_uri can't be nil or an empty string" if
+        uri == '' || uri.nil? == :true
+      fail 'source_uri should be of the format <media>:<filename>' unless
+        uri.include?(':')
+    end
+    munge do |uri|
+      image = {}
+      if uri.include?('/')
+        image[:media] = uri.split('/')[0]
```
mikewiebe: This is a case where I think some comments showing some sample URI formats would be good.

lib/puppet/type/cisco_upgrade.rb
```diff
@@ -0,0 +1,123 @@
...
+      image
+    end
+  end # param source_uri
+
+  newparam(:force_upgrade) do
+    desc 'Force upgrade the device.'
+    defaultto :false
+    newvalues(:true, :false)
+    munge do |value|
+      value = true if value == :true
```
mikewiebe: Two things.  Here for parameter `force_upgrade` and below for `delete_boot_image` add a comment indicating why the munge is needed and you can simplify to the following:

`munge { |value| value == :true ? true : false }`

mikewiebe:
@rahushen Looks good.  Just a few minor comments.

lib/puppet/type/cisco_upgrade.rb
mikewiebe: Update all references to `media:image` to be `uri:image` instead.

lib/puppet/type/cisco_upgrade.rb
```diff
@@ -0,0 +1,123 @@
...
+    ..attributes..
+  }
+  ```
+
+  There can only be one instance of cisco_upgrade i.e. 'image'
+
+  Example:
+  ```
+    cisco_upgrade {'image' :
+      version           => '7.0(3)I5(1)'
```
mikewiebe: Add a `,` at the end of the version line.  I cut and pasted this into my manifest and found it :)

lib/puppet/provider/cisco_upgrade/cisco.rb
```diff
@@ -0,0 +1,83 @@
...
+    inst = []
+    service = Cisco::Service
+
+    inst << new(
+      name:              'image',
+      version:           service.image_version,
+      source_uri:        service.image,
+      force_upgrade:     :false,
+      delete_boot_image: :false)
+  end
```
mikewiebe: You don't need to include `source_uri:`, `force_upgrade:` or `delete_boot_image:` as part of inst.  Only need to include properties, not parameters.

mikewiebe:
@rahushen Looks good.  Just one more minor comment and we should be able to merge.

lib/puppet/type/cisco_upgrade.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  newparam(:name, namevar: :true) do
+    # Parameter used only to satisfy namevar
+    desc 'Name of cisco_upgrade instance. Valid values are string'
+    validate do |name|
+      warning "only 'image' is accepted as a valid name" if name != 'image'
+    end
+  end
+
+  newparam(:source_uri) do
```
mikewiebe: Just realized that we need to make `:source_uri` mandatory.  If the manifest does not specify the parameter or if the `puppet resource cisco_upgrade image version=ver` command does not specify this parameter we get a very cryptic error.

Use a validate block at the end of the type to do a nil check and fail indicating the parameter is mandatory:

Example: https://github.com/cisco/cisco-network-puppet-module/blob/2f1d60e358b7009bfedcf6175357603c9e115831/lib/puppet/type/cisco_bgp_neighbor_af.rb#L490

mikewiebe:
👍

PR #429, opened 2017-03-06
Feature/upgrade services

rahushen:
Moved validation code for source_uri from 'type' to 'provider'


lib/puppet/provider/cisco_upgrade/cisco.rb
```diff
@@ -71,6 +71,8 @@ def version=(new_version)
     return if new_version.nil?
     # Convert del_boot_image and force_upgrade from symbols
     # to Boolean Class
+    fail 'The source_uri parameter must be set in the manifest' if
+      @resource[:source_uri].nil?
```
mikewiebe: Should you also check for empty string here?

rahushen: source_uri will be a hash at this point because of the munging done in the type

mikewiebe:
👍

PR #430, opened 2017-03-06
Feature/upgrade services

rahushen:
Added Beaker tests


tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -75,6 +75,7 @@ def hash_to_patterns(hash)
     if /^\[.*\]$/.match(value)
       value.gsub!(/[\[\]]/) { |s| '\\' + "#{s}" }.gsub!(/\"/) { |_s| '\'' }
     end
+    value.gsub!(/[\(\)]/) { |s| '\\' + "#{s}" } if /\(.*\)/.match(value)
```
mikewiebe: What was the reason for this change?

tests/beaker_tests/cisco_upgrade/test_upgrade_idempotence.rb
```diff
@@ -0,0 +1,59 @@
...
+}
+
+# Non-default Tests. NOTE: [:resource] = [:manifest_props] for all non-default
+tests[:non_default] = {
+  desc:           '1.1 Non_Defaults',
+  code:           [0],
+  title_pattern:  'image',
+  manifest_props: {
+    version:           @version,
+    source_uri:        on(agent, facter_cmd('-p cisco.images.system_image')).output,
```
mikewiebe: What happens if the `source_uri:` is something other then `bootflash:`

tests/beaker_tests/cisco_upgrade/test_upgrade_idempotence.rb
```diff
@@ -0,0 +1,59 @@
...
+#  - A description of the 'tests' hash and its usage
+#
+###############################################################################
+require File.expand_path('../../lib/utilitylib.rb', __FILE__)
+@version = image_version.to_s.strip
+# Test hash top-level keys
+tests = {
+  master:           master,
+  agent:            agent,
+  operating_system: 'nexus',
```
mikewiebe: Limit this to only run on N3k and N9k platforms.

https://github.com/cisco/cisco-network-puppet-module/blob/develop/tests/beaker_tests/cisco_evpn_vni/test_evpn_vni.rb#L30

tests/beaker_tests/lib/utilitylib.rb
rahushen: The version string is usually something of the sort '7.0(3)I2(5)'. The parenthesis in the string will cause Match of the generated regexp to fail. Hence need to escape those characters.

tests/beaker_tests/cisco_upgrade/test_upgrade_idempotence.rb
rahushen: The test would most likely fail because node-utils would return fail. We could skip the test if image not on bootflash: , but in a controlled test environment, I'm not sure if we're going to run into this.

tests/beaker_tests/cisco_upgrade/test_upgrade_idempotence.rb
rahushen: Done

tests/beaker_tests/lib/utilitylib.rb
mikewiebe: My concern is that this change will break other tests that use this method.  Probably need to to run a full beaker test to make sure.

tests/beaker_tests/cisco_upgrade/test_upgrade_idempotence.rb
```diff
@@ -0,0 +1,60 @@
...
+
+# Non-default Tests. NOTE: [:resource] = [:manifest_props] for all non-default
+tests[:non_default] = {
+  desc:           '1.1 Non_Defaults',
+  code:           [0],
+  title_pattern:  'image',
+  platform:       'n(3|9)k',
+  manifest_props: {
+    version:           @version,
+    source_uri:        on(agent, facter_cmd('-p cisco.images.system_image')).output,
```
mikewiebe: What happens if `source_uri:` is something other then `bootflash:`

rahushen: The test would most likely fail because node-utils would return fail. We could skip the test if image not on bootflash: , but in a controlled test environment, I'm not sure if we're going to run into this.

tests/beaker_tests/cisco_upgrade/test_upgrade_idempotence.rb
rahushen: I've kickstarted a full beaker test run. Will update with results

tests/beaker_tests/lib/utilitylib.rb
rahushen: I've kickstarted a full beaker test run. Will update with the results.

README.md
```diff
@@ -4011,6 +4014,50 @@ Specifies a preshared key for the host. Valid values are 'clear', 'encrypted',
...
+Manages the upgrade of a Cisco device.
+
+| Platform | OS Minimum Version | Module Minimum Version |
+|----------|:------------------:|:----------------------:|
+| N9k      | 7.0(3)I2(1)        | 1.6.0                  |
+| N3k      | 7.0(3)I2(1)        | 1.6.0                  |
+| N5k      | not applicable     | not applicable         |
+| N6k      | not applicable     | not applicable         |
+| N7k      | not applicable     | not applicable         |
+| N9k-F    | not applicable     | not applicable         |
```
mikewiebe: We should support this on n9k fretta.  We will need to verify, but it should just work.

README.md
```diff
@@ -4011,6 +4014,50 @@ Specifies a preshared key for the host. Valid values are 'clear', 'encrypted',
 "Specifies the preshared key password for the host. Valid value is a string.
 
 --
+### Type: cisco_upgrade
+
+Manages the upgrade of a Cisco device.
+
+| Platform | OS Minimum Version | Module Minimum Version |
+|----------|:------------------:|:----------------------:|
+| N9k      | 7.0(3)I2(1)        | 1.6.0                  |
```
mikewiebe: I would specify this as `7.0(3)I2(2e)`  No reason to go earlier for a new resource.

README.md
```diff
@@ -4011,6 +4014,50 @@ Specifies a preshared key for the host. Valid values are 'clear', 'encrypted',
...
+| `source_uri`    | Only images on `bootflash:`  are supported. |
+
+#### Parameters
+
+##### `name`
+Name of cisco_upgrade instance. Valid values are string.
+*Only 'image' is a valid name for the cisco_upgrade resource.*
+
+##### `source_uri`
+URI to the image to install on the device. Format <uri>:<image>.Valid values are string.
```
mikewiebe: Perhaps `Image upgrade uri` is simpler?

README.md
```diff
@@ -4011,6 +4014,50 @@ Specifies a preshared key for the host. Valid values are 'clear', 'encrypted',
...
+
+#### <a name="cisco_upgrade-caveats">Caveats</a>
+
+The `cisco_upgrade` is only supported on *simplex* N3K and N9K devices. HA devices are currently not supported. 
+
+| Property | Caveat Description |
+|:--------|:-------------|
+| `source_uri`    | Only images on `bootflash:`  are supported. |
+
+#### Parameters
```
mikewiebe: Somewhere in here, it would be good to add a note indicating that the puppet file provider can be used to manage image files on bootflash: along with a pointer to the sample manifest.

README.md
mikewiebe: The format did not render properly.

tests/beaker_tests/cisco_upgrade/test_upgrade_idempotence.rb
mikewiebe: Fair enough.

examples/demo_all_cisco.pp
```diff
@@ -49,6 +49,7 @@
   include ciscopuppet::cisco::demo_stp_vlan
   include ciscopuppet::cisco::demo_tacacs_server
   include ciscopuppet::cisco::demo_tacacs_server_host
+  include ciscopuppet::cisco::demo_upgrade
```
mikewiebe: I would comment this out in the all file so that it does not get run by default.

mikewiebe:
👍

PR #431, opened 2017-03-07
remove all route maps after beaker test

saichint:
This PR is for removing all route maps before/after the beaker tests so that no residual configuration is left on the system

### DIFF
--- a/tests/beaker_tests/cisco_route_map/test_route_map.rb
+++ b/tests/beaker_tests/cisco_route_map/test_route_map.rb
@@ -540,11 +540,7 @@ def unsupported_properties(_tests, _id)
 end
 
 def cleanup(agent)
-  test_set(agent, 'no route-map rm1')
-  test_set(agent, 'no route-map rm2')
-  test_set(agent, 'no route-map rm3')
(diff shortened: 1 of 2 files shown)


tests/beaker_tests/cisco_route_map/test_route_map.rb
```diff
@@ -540,11 +540,7 @@ def unsupported_properties(_tests, _id)
 end
 
 def cleanup(agent)
-  test_set(agent, 'no route-map rm1')
-  test_set(agent, 'no route-map rm2')
-  test_set(agent, 'no route-map rm3')
-  test_set(agent, 'no route-map rm4')
-  test_set(agent, 'no route-map rm5')
+  remove_all_route_maps(agent)
```
mikewiebe: Instead of creating your own remove method, can you use the one we already have?

https://github.com/cisco/cisco-network-puppet-module/blob/develop/tests/beaker_tests/cisco_bgp_af/test_bgpaf.rb#L285

saichint:
sorry, my mistake. I copied remove_all_vrfs methodology. There was no need for that

PR #438, opened 2017-03-23
Replace heavy_minus_sign emoji with utf-8 ➖symbol in README

shermdog:
Previously the README was a mix of utf-8 symbols and GitHub emoji.  This was not properly rendering on Puppet Forge, as GitHub emoji codes are not supported.


mikewiebe:
@shermdog Thanks for fixing this! 👍

Issue #424, opened 2017-02-15
ensure => absent on Ethernet interfaces fails


tomcooperca:
Hey all,
I'm running through some basic manifests for testing & validation of the module against our config. I have a manifest for each corresponding feature that applies some basic dummy configuration then a removal class that cleans it up.

* Cisco Nexus N9K-C93180YC-EX running NX-OS 7.0(3)I4(2)
* ciscopuppet module & cisco_node_utils version 1.5.0
* Puppet version 4.7.0 (PE 2016.4.2)

In this manifest, I have two Ethernet interfaces to become part of a port-channel, which then is configured as a vPC Peer Link. However, when I run through the removal class to un-configure the Ethernet interfaces (using `cisco_interface { 'Ethernet1/10': ensure => absent, }` ), a CLI error is produced. See below:

**vpc.pp**
```
...<truncated>
  cisco_interface { "port-channel${peerlink_po_id}" :
    ensure                  => present,
    description             => 'vPC Peerlink',
    switchport_mode         => 'trunk',
    vpc_peer_link           => true,
    stp_port_type           => 'network',
    storm_control_broadcast => '90.00',
    storm_control_multicast => '90.00',
    shutdown                => false,
    require                 => Cisco_vpc_domain["${domain}"],
  }
  $peerlink_interfaces.each |String $interface| {
    cisco_interface { $interface :
      ensure                  => present,
      description             => "vPC Peerlink Po${peerlink_po_id}",
      switchport_mode         => 'trunk',
      stp_port_type           => 'network',
      storm_control_broadcast => '90.00',
      storm_control_multicast => '90.00',
      shutdown                => false,
    }
    cisco_interface_channel_group { "${interface}-Po${peerlink_po_id}" :
      ensure        => present,
      interface     => $interface,
      channel_group => $peerlink_po_id,
      require       => Cisco_interface["port-channel${peerlink_po_id}"],
    }
  }
...
```

**vpc_removal.pp**
```
...<truncated>
  $peerlink_interfaces.each |String $interface| {
    cisco_interface_channel_group { "${interface}-Po${peerlink_po_id}" :
      ensure => absent,
    }
    cisco_interface { $interface :
      ensure => absent,
    }
  }
```

Here's the `--debug` run while trying to run the vpc_removal class:

```
Notice: /Stage[main]/Profile::Cisco::Testcase::Vpc_remove/Cisco_interface[Ethernet1/10]/ensure: removed
[...]

PR #444, opened 2017-04-07
Eth int default

saichint:
This PR for adding purge_config property to interface provider so that the physical interface can be put back into default state. All tests pass.
 This update fixes the issue with the cisco_interface provider described in https://github.com/cisco/cisco-network-puppet-module/issues/424


saichint:
Need to add beaker test, closing it for now, will reopen after adding it

PR #445, opened 2017-04-07
purge_config add to interface

saichint:
This PR for adding purge_config property to interface provider so that the physical interface can be put back into default state. All tests pass on n3k, n6k, n9k.
This update fixes the issue with the cisco_interface provider described in https://github.com/cisco/cisco-network-puppet-module/issues/424

### DIFF
--- a/lib/puppet/type/cisco_interface.rb
+++ b/lib/puppet/type/cisco_interface.rb
@@ -1279,4 +1279,14 @@ Puppet::Type.newtype(:cisco_interface) do
 
     munge { |value| value == 'default' ? :default : Integer(value) }
   end # property load_interval_counter_3_delay
+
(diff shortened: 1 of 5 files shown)


lib/puppet/type/cisco_interface.rb
```diff
@@ -1279,4 +1279,14 @@ def is_to_s(value)
...
+
+  newproperty(:purge_config) do
+    desc 'Puts the ethernet interface in default state.'
+
+    newvalues(:true)
+  end
+
+  validate do
+    fail ArgumentError, 'All params MUST be nil if purge_config is true' if self[:purge_config] == :true && properties.length > 2
+  end
```
saichint: added

PR #448, opened 2017-04-17
Replace emoji heavy_minus_sign with actual utf8 character in README.

scotje:
Since the checkboxes are already utf-8 characters, switching the heavy minus signs should make it render consistently everywhere.


mikewiebe:
@scotje I merged the following PR a few weeks back that made the same change:

https://github.com/cisco/cisco-network-puppet-module/pull/438/files

scotje:
Oh cool! Nevermind then!

PR #461, opened 2017-07-06
Add separate test for match_src_proto

rahushen:
Just like the minitest, the beaker tests for route map need to be updated for handling the unordered representation of match_src_proto in running config on I7 (n9k only).


saichint:
:+1:

PR #462, opened 2017-08-04
Puppet5 test fix

rahushen:
This PR address the line wrap change that is seen in Puppet5. Changes made will be tested to retain compatibility with the existing Puppet version

Test Results :

Puppet 5 - 102P / 2 F / 11 S / 2 E
Puppet (current version) - 102/ 3F / 11S / 1E

2 of the failure are similar and needs to be investigated separately. The vrf testcases that errored out need to be investigated separately and they are not related to this diff.


saichint:
:+1:

PR #463, opened 2017-08-11
Fix for CISCO-50, CISCO-51, CISCO-52

saichint:
This PR is for fixing 3 customer reported issues:
1. snmp user throws error that user "admin" cannot be deleted
2. aaa_authorization_login_cfg_svc and aaa_authorization_login_exec_svc are not idempotent for groups property.
3. snmp_user 'admin' edge case

### DIFF
--- a/lib/puppet/provider/cisco_snmp_user/cisco.rb
+++ b/lib/puppet/provider/cisco_snmp_user/cisco.rb
@@ -214,7 +214,7 @@ Puppet::Type.type(:cisco_snmp_user).provide(:cisco) do
   end
 
   def unconfigure_snmp_user
-    @snmp_user.destroy
+    @snmp_user.destroy unless @resource[:user] == 'admin'
     @snmp_user = nil
     @property_hash[:ensure] = :absent
   end
--- a/lib/puppet/provider/cisco_snmp_user/cisco.rb
+++ b/lib/puppet/provider/cisco_snmp_user/cisco.rb
@@ -224,6 +224,9 @@ Puppet::Type.type(:cisco_snmp_user).provide(:cisco) do
 
   def flush
     if @property_flush[:ensure] == :absent
+      fail ArgumentError,
+           'The admin account cannot be deactivated on this platform. ' \
+            if @resource[:user] == 'admin'
       unconfigure_snmp_user
     elsif @property_flush[:ensure] == :present
       configure_snmp_user
--- a/lib/puppet/provider/cisco_snmp_user/cisco.rb
+++ b/lib/puppet/provider/cisco_snmp_user/cisco.rb
@@ -214,6 +214,9 @@ Puppet::Type.type(:cisco_snmp_user).provide(:cisco) do
   end
 
   def unconfigure_snmp_user
+    # admin user is always present on the device and it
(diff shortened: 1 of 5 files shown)


lib/puppet/provider/cisco_snmp_user/cisco.rb
```diff
@@ -214,7 +214,7 @@ def handle_attribute_update
   end
 
   def unconfigure_snmp_user
-    @snmp_user.destroy
+    @snmp_user.destroy unless @resource[:user] == 'admin'
```
mikewiebe: Please add a comment for future developers to indicate why we don't call destroy for the `admin` user.

lib/puppet/type/cisco_ace.rb
```diff
@@ -291,4 +291,12 @@ def name
     desc 'Log matches against this entry'
     newvalues(:true, :false)
   end
+
+  validate do
+    unless self[:remark].nil?
+      fail ArgumentError,
+           'established and log parameters must be nil when remark is not nil' unless
```
mikewiebe: *nit:* I would modify this to read:
`'established' and 'log' properties should not be set for `remark` ace.`

mikewiebe: @shermdog Interested in your opinion here.  The only reason we are adding this validate check is because prior to this change the output of the `puppet resource` command would display these two properties set to `false` even though they are not used for the remark ace.  Older manifests will error out with a strange error message so this validate provides more meaningful guidance.

shermdog:
For the snmp_user I would suggest validating the resource rather than silently ignoring the change

```

**Comment by shermdog** [comment]
For the cisco_snmp_user I suggest doing some validation instead of skipping, as the information passed back to the user is incorrect:
```
Notice: /Stage[main]/Main/Cisco_snmp_user[admin]/ensure: removed
```

Something like this is better
```ruby
  def destroy
    fail ArgumentError,
         "The admin account cannot be deactivated on this platform. " \
            if @resource.name == 'admin'
    info "Removing snmp user #{@property_hash[:name]}"
    @property_flush[:ensure] = :absent
  end
```
```
Error: Could not set 'absent' on ensure: The admin account cannot be deactivated on this platform.  at /root/snmp.pp:1
Error: Could not set 'absent' on ensure: The admin account cannot be deactivated on this platform.  at /root/snmp.pp:1
Wrapped exception:
The admin account cannot be deactivated on this platform. 
Error: /Stage[main]/Main/Cisco_snmp_user[admin]/ensure: change from present to absent failed: Could not set 'absent' on ensure: The admin account cannot be deactivated on this platform.  at /root/snmp.pp:1
```

I'm also not super stoked about the idea of the provider deleting the user to make changes.  It appears the system allows for changing the existing resource.  Multiple get/set really kills are performance.

This work also needs to be applied to the snmp_user (netdev) providers.

(I'll review other changes next).

shermdog:
Everything seems to be good.  I would like to see tests added to ensure regressions aren't introduced later.  @mikewiebe let's talk about the destroy/create bit and multiple get/sets.

mikewiebe:
@shermdog I am going to go ahead and merge this but we will create new user stories for additional work/refactoring after our discussion tomorrow.

PR #464, opened 2017-08-18
Puppet5 fixes

rahushen:
Made fixes to issues seen in puppet5. Errors seen in the tests were because `remove_all_vrfs` was silently failing to do it's job. Should now be fixed. 

All failed/errored tests were retested with puppet5 and latest puppet and are passing.

### DIFF
--- a/tests/beaker_tests/cisco_command_config/test_command_config.rb
+++ b/tests/beaker_tests/cisco_command_config/test_command_config.rb
@@ -192,6 +192,19 @@ def test_set_get
 
   logger.info('* check config')
   on(agent, cmd_prefix + "test_get='incl loopback1'")
+  # The output of test_get has changed in Puppet5 and newer versions of Puppet.
+  # Old output:
+  # cisco_command_config { 'cc':
+  #   test_get => '
+  # interface loopback1
+  # interface loopback10
+  # ',
+  # }
+  # New output:
+  # cisco_command_config { 'cc':
+  #   test_get => "\ninterface loopback1\ninterface loopback10\n",
+  # }
+  # Modifying the below regular expression to make ^ and \n optional.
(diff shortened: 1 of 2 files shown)


tests/beaker_tests/cisco_command_config/test_command_config.rb
```diff
@@ -193,7 +193,7 @@ def test_set_get
   logger.info('* check config')
   on(agent, cmd_prefix + "test_get='incl loopback1'")
   fail_test("TestStep :: set/get :: FAIL\nstdout:\n#{stdout}") unless
-    stdout[/^interface loopback1/]
+    stdout[/^?\n?interface loopback1/]
```
mikewiebe: Can you add some sample output from both agent versions that this is trying to match?  Also add comments to make it clear that we need to match two versions of output.

tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -1463,8 +1463,8 @@ def remove_all_vlans(agent, stepinfo='Remove all vlans & bridge-domains')
 end
 
 def remove_all_vrfs(agent)
-  found = test_get(agent, "incl 'vrf context' | excl management").split("\n")
-  found.map! { |cmd| "no #{cmd}" if cmd[/^vrf context/] }
+  found = test_get(agent, "incl 'vrf context' | excl management").split('\\n')
```
mikewiebe: Same comment here.

tests/beaker_tests/cisco_command_config/test_command_config.rb
rahushen: Done.

tests/beaker_tests/lib/utilitylib.rb
rahushen: Done.

mikewiebe:
👍

PR #467, opened 2017-09-22
bridge_domain fixes

saichint:
This PR is for fixing bridge-domain name default action and also adding 'default' keyword to fabric_control and shutdown properties. This is missing in the current code.


mikewiebe:
@saichint Changes look ok but do we need to add test coverage?

mikewiebe:
👍  Based on our discussion.

PR #469, opened 2017-09-22
remove peerlist and ingress replication for n7k

saichint:
This PR is for removing peer-list and ingress replication from n7k as they are not supported on that platform. Tests pass


mikewiebe:
👍

PR #477, opened 2017-11-27
Use correct host_reachability flood for vxlan_vtp_vni dependencies

mikewiebe:
Fixes the test to use the correct and supported option.


saichint:
:+1:

PR #478, opened 2017-11-28
Admin cookie override

mikewiebe:
This update adds documentation for running puppet on Nexus with a user other then `admin`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
 
 1. [Module Description](#module-description)
 1. [Setup](#setup)
+   * [Puppet Master](#setup-puppet-master)
+   * [Puppet Agent](#setup-puppet-agent)
 1. [Example Manifests](#example-manifests)
 1. [Resource Reference](#resource-reference)
    * [Resource Type Catalog (by Technology)](#resource-by-tech)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -94,6 +97,44 @@ Once installed, the GEM will remain persistent across system reloads within the
...
+
+First create a different user with the role `network-admin`.
+
+~~~
+config term
+  username puppetuser password puppet role network-admin
+end
+~~~
+
+Next create a file called `cisco_node_utils.yaml` under the `cisco-network-puppet-module` files directory and add a cookie `puppetuser:local` under the `default:` yaml key.
```
saichint: I am a bit confused by the comment here. It says, create file under 'cisco-network-puppet-module' directory but it is in  ciscopuppet/files directory. Not exactly sure how to correlate these two. Is it assumed that 'cisco-network-puppet-module' directory means '/etc/puppetlabs/code/environments/production/modules/ciscopuppet/files' ?

mikewiebe: I will update the comment to clarify.  I can see how it would be confusing.

mikewiebe:
@saichint Comment addressed.

saichint:
:+1:

PR #479, opened 2017-12-07
(CISCO-60) tacacs_global timeout undef

shermdog:
This commit changes the output of timeout from -1 to undef when not set.  This more accurately reflects the real world state of the resource.  Also allows users to set undef in their manifest to accept default value when configured.


mikewiebe:
Looks good but can you add some documentation [here](https://github.com/cisco/cisco-network-puppet-module#type-tacacs_global) around that fact that users can now use `undef`.

shermdog:
@mikewiebe added.

PR #483, opened 2018-02-06
TRM (part 1)

rahushen:
- added type, provider and beaker tests for evpn_multicast
- added type, provider and beaker tests for cisco_vrf_af modifications

### DIFF
--- a/examples/cisco/demo_trm.pp
+++ b/examples/cisco/demo_trm.pp
@@ -16,38 +16,42 @@
 
 class ciscopuppet::cisco::demo_trm {
 
-  if platform_get() =~ /n9k/ {
+  if platform_get() =~ /n9k(-ex)?$/ {
 
-    $mvpn_support = prop_supported('route_target_import_mvpn')
+    if image_supports_trm() {
 
-    $rt_both_mvpn = $mvpn_support ? {
-      true    => true,
-      default => undef
+      $mvpn_support = prop_supported('route_target_import_mvpn')
+
+      $rt_both_mvpn = $mvpn_support ? {
+        true    => true,
+        default => undef
+      }
+
+      $rt_import_mvpn = $mvpn_support ? {
+        true    => ['12:13', '14:15'],
+        default => undef
+      }
+
+      $rt_export_mvpn = $mvpn_support ? {
+        true    => ['16:17', '18:19'],
+        default => undef
+      }
+
+      cisco_evpn_multicast { 'default':
+        ensure          => present,
+      }
+
+      cisco_vrf_af { 'red ipv4 unicast':
+        ensure                        => present,
+        route_policy_export           => 'abc',
+        route_policy_import           => 'abc',
+        route_target_import_mvpn      => $rt_import_mvpn,
+        route_target_export_mvpn      => $rt_export_mvpn,
+        route_target_both_auto_mvpn   => $rt_both_mvpn,
+      }
+    } else {
+      notify{'SKIP: This image does not support TRM': }
     }
-
-    $rt_import_mvpn = $mvpn_support ? {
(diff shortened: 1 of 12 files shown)


examples/cisco/demo_trm.pp
```diff
@@ -0,0 +1,54 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+class ciscopuppet::cisco::demo_trm {
+
+  if platform_get() =~ /n9k/ {
```
mikewiebe: Not clear on how this indicates the platform supports TRM?

lib/puppet/provider/cisco_evpn_multicast/cisco.rb
```diff
@@ -0,0 +1,69 @@
...
+    @property_flush[:ensure] = :present
+  end
+
+  def destroy
+    @property_flush[:ensure] = :absent
+  end
+
+  def flush
+    if @property_flush[:ensure] == :absent
+      @nu = Cisco::EvpnMulticast.new
```
mikewiebe: I believe you can avoid creating the object here just to destroy it by moving the call `@nu = Cisco::EvpnMulticast.new` into the initialize method.

Example: https://github.com/cisco/cisco-network-puppet-module/blob/develop/lib/puppet/provider/cisco_bfd_global/cisco.rb#L65

tests/beaker_tests/cisco_vrf_af/test_vrf_af.rb
```diff
@@ -37,22 +37,28 @@
...
   resource:       {
     'route_target_both_auto'      => 'false',
     'route_target_both_auto_evpn' => 'false',
+    'route_target_both_auto_mvpn' => 'false',
   },
 }
 
 routetargetimport          = ['1.1.1.1:55', '1:33']
 routetargetimportevpn      = ['2.2.2.2:55', '2:33']
+routetargetimportmvpn      = ['7.7.7.7:55', '7:33']
```
mikewiebe: We are sorting this list in both the type and node_utils right?  Just making sure we will be idempotent regardless of what order these are specified in.

mikewiebe:
@rahushen Done with review.

examples/cisco/demo_trm.pp
rahushen: I've updated this to add another check for image as well.

lib/puppet/provider/cisco_evpn_multicast/cisco.rb
rahushen: @mikewiebe .. I removed that because of the way I've written the NU piece for this. I call create (in node utils) every single time and this was causing it causing evpn_multicast to be turned on while doing 'puppet resource'. Since this is really an ensure present/absent kind of provider I moved the nu object creation to flush.

tests/beaker_tests/cisco_vrf_af/test_vrf_af.rb
rahushen: Yes, this works even with the list reversed and idempotence is maintained.

lib/puppet/provider/cisco_evpn_multicast/cisco.rb
mikewiebe: Gotcha.  Maybe add a clarification comment.

rahushen: Done.

rahushen:
@mikewiebe I have made the recommended changes

PR #484, opened 2018-02-14
TRM: IP multicast provider, type and beaker testcases.

rahushen:


examples/cisco/demo_trm.pp
```diff
@@ -49,6 +49,12 @@
         route_target_export_mvpn      => $rt_export_mvpn,
         route_target_both_auto_mvpn   => $rt_both_mvpn,
       }
+
+      cisco_ip_multicast { 'default':
+        ensure =>                 present,
+        overlay_distributed_dr => 'true',
+        overlay_spt_only =>       'true',
```
mikewiebe: Make sure to enable `nv overlay evpn` in the node_utils for `overlay_spt_only`

lib/puppet/type/cisco_ip_multicast.rb
```diff
@@ -0,0 +1,85 @@
...
+  # Parameters #
+  ##############
+  newparam(:name, namevar: true) do
+    # Creating a parameter to satisfy namevar condition
+    # Only 'default' is an accepted value
+    desc "Instance of IP Multicast, only allow the value 'default'"
+
+    validate do |name|
+      if name != 'default'
+        error "only 'default' is accepted as a valid ip_multicast resource name"
```
mikewiebe: *nit* I would prepend ip_multicast with cisco.
`error "only 'default' is accepted as a valid cisco_ip_multicast resource name"`

tests/beaker_tests/cisco_ip_multicast/test_ip_multicast.rb
```diff
@@ -0,0 +1,90 @@
...
+
+# Test hash test cases
+tests[:default] = {
+  desc:           '1.1 Default Properties',
+  title_pattern:  'default',
+  manifest_props: {
+    overlay_distributed_dr: 'default',
+    overlay_spt_only:       'default',
+  },
+  resource:       {}, # no output expected when all defaults
```
mikewiebe: If the manifest uses `ensure => present` then I would expect the resource to display both properties set to the default value of `false`.

examples/cisco/demo_trm.pp
rahushen: Done.

lib/puppet/type/cisco_ip_multicast.rb
rahushen: done

tests/beaker_tests/cisco_ip_multicast/test_ip_multicast.rb
rahushen: done

mikewiebe:
👍  Will merge when travis completes.

PR #486, opened 2018-05-10
Remove host_reachability from dependency_manifest call

mikewiebe:
Setting the `host_reachability` in the dependency manifest call impacts tests since some versions of Nexus require it to be set properly for ingress_replication.  We let the setter for ingress replication set the value properly.

Tests pass on N9k (h_dev, greensboro) and N7k.


saichint:
👍

PR #487, opened 2018-05-31
(CISCO-66) add missing confine to netdev_stdlib providers

shermdog:
The following providers were missing their confine.  This module supports netdev_stdlib and has to play nice with other providers.
A confine helps ensure this module does not interfere with other providers.

* network_trunk
* network_vlan
* port_channel


mikewiebe:
👍

PR #490, opened 2018-07-09
Nexus release 9.2 documentation updates.

mikewiebe:
**Summary:**
This PR updates documentation for the following:
  * Indicate that puppet agent installs into the `bash-shell `are no longer supported.
  * Add n3k low footprint workflow when using `guestshell`


saichint:
👍

PR #491, opened 2018-07-10
Fix CISCO-67

saichint:
This PR fixes issue (https://tickets.puppetlabs.com/browse/CISCO-67)
If tacacs_server ```encryption_password``` is provided but not ```encryption_type```, it should be error out. 

Add check to see if the type is ```nil``` or ```none``` and throw error if password is present is the manifest.

### DIFF
--- a/lib/puppet/provider/cisco_tacacs_server/cisco.rb
+++ b/lib/puppet/provider/cisco_tacacs_server/cisco.rb
@@ -1,7 +1,7 @@
 ################################################
 # January, 2015
 #
-# Copyright (c) 2015-2016 Cisco and/or its affiliates.
+# Copyright (c) 2015-2018 Cisco and/or its affiliates.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -17,6 +17,13 @@
 #################################################
 
 require 'cisco_node_utils' if Puppet.features.cisco_node_utils?
+begin
+  require 'puppet_x/cisco/cmnutils'
+rescue LoadError # seen on master, not on agent
+  # See longstanding Puppet issues #4248, #7316, #14073, #14149, etc. Ugh.
+  require File.expand_path(File.join(File.dirname(__FILE__), '..', '..', '..',
+                                     'puppet_x', 'cisco', 'cmnutils.rb'))
+end
 
 Puppet::Type.type(:cisco_tacacs_server).provide(:cisco) do
   desc 'The Cisco provider.'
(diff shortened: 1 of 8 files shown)


mikewiebe:
👍 @shermdog Can you please review?

shermdog:
This looks good, but I did uncover the resource does not return the `encryption_type` after it's been set.

```
[root@puppet-device-devel ~]# puppet device -v -t nxapi --resource cisco_tacacs_server --trace
Info: retrieving resource: cisco_tacacs_server from nxapi at file:///root/nxapi.yaml
cisco_tacacs_server { 'default':
  ensure              => 'present',
  deadtime            => 0,
  directed_request    => 'false',
  encryption_password => '"wwgfus-pwb"',
  source_interface    => 'mgmt0',
  timeout             => 7,
}
```
tacacs_global returns the key type
```
[root@puppet-device-devel ~]# puppet device -v -t nxapi --resource tacacs_global --trace
Info: retrieving resource: tacacs_global from nxapi at file:///root/nxapi.yaml
tacacs_global { 'default':
  key              => '"wwgfus-pwb"',
  key_format       => 7,
  source_interface => ['mgmt0'],
  timeout          => 7,
}
```
```
jleb8uc5m2lp77z# sh run | inc tacacs
feature tacacs+
tacacs-server key 7 "wwgfus-pwb"
```

saichint:
@shermdog @mikewiebe  I can take a look at this but please remember that the encryption_type returned will always be 7, irrespective of what type user has given in the manifest. This is because the running config cannot show any clear text passwords so it will show only encrypted passwords and type will be 7. So this is almost like hardcoding the type to 7 in the resource. It does not add any value other than keeping it consistent with other resources.

shermdog:
@saichint let's return the encryption_type when the key is set.  This will ensure that users are able to directly use `puppet resource` output in their manifests.

saichint:
@shermdog @mikewiebe Fixed the puppet resource issue for tacacs_server, however found similar issues with tacacs_server_host as well. When fixing this module, moved the code to common utiltities to reduce code repetition.

shermdog:
This is working great for apply and resource, but causing issues when used with a master

```
Linux# puppet agent -tv        
Info: Using configured environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Retrieving locales
Info: Loading facts
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Could not autoload puppet/type/cisco_tacacs_server_host: Could not autoload puppet/provider/cisco_tacacs_server_host/cisco: uninitialized constant PuppetX::NUConst::Cisco (file: /etc/puppetlabs/code/environments/production/manifests/site.pp, line: 45, column: 1) on node cfl9wwivwuet42g.delivery.puppetlabs.net
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

The issue goes away on subsequent runs, which is related to how puppet falsely ignores subsequent load errors.  

We've seen this in other modules, which we generally get around by using `require_relative`.  I haven't gone too deep down this particular rabbit hole to figure out the root cause here.

saichint:
@shermdog I was able to reproduce the error you mentioned. 
Moved the new constants inside puppetx::cisco module. Retested and it is working good. 
Please check it again and merge if you do not have any more comments.

shermdog:
@saichint I updated the beaker tests for this and added a commit here.  I'm good w/ the merge, but figured you or @mikewiebe would want to put eyes on the test update first.

saichint:
@shermdog Have you run the beaker with these updated tests? The changes look ok to me. @mikewiebe also can check it and leave any comments.

shermdog:
I ran this through beaker manually, but I can run this branch through full CI.

mikewiebe:
👍 Tests look pretty straight forward to me.

saichint:
@shermdog can you merge if everything is fine?

PR #492, opened 2018-07-17
(maint) fix cisco_ip_multicast test defaults

shermdog:
The default value for overlay_spt_only flipped between I7.3 and I7.4
This commit addresses this behavior, which was already fixed in node_utils cisco/cisco-network-node-utils#582

### DIFF
--- a/tests/beaker_tests/cisco_ip_multicast/test_ip_multicast.rb
+++ b/tests/beaker_tests/cisco_ip_multicast/test_ip_multicast.rb
@@ -37,6 +37,9 @@ skip_unless_supported(tests)
 # Skip -ALL- tests if feature is not supported by image version
 skip_nexus_image(/I[2-6]/, tests)
 
+# Default values changed between I7.3 and I7.4
+spt_default = nexus_image[/7\.0\(3\)I7\([1-3]\)/] ? false : true
+
 # Test hash test cases
 tests[:default] = {
   desc:           '1.1 Default Properties',
@@ -47,7 +50,7 @@ tests[:default] = {
   },
   resource:       {
     overlay_distributed_dr: 'false',
-    overlay_spt_only:       'false',
+    overlay_spt_only:       spt_default.to_s,
   },
 }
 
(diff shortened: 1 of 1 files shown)


saichint:
👍

tests/beaker_tests/cisco_ip_multicast/test_ip_multicast.rb
```diff
@@ -37,6 +37,9 @@
 # Skip -ALL- tests if feature is not supported by image version
 skip_nexus_image(/I[2-6]/, tests)
 
+# Default values changed between I7.3 and I7.4
```
mikewiebe: Nice!  I was literally setting up my environment to work on this issue and saw your PR.

This change in default behavior is also present in our latest Hamilton release which is `NXOS: version 9.2(1)` so this should handle that case as well.

mikewiebe: Thinking more, this will not handle our older releases.  It should also be false for

`7.0(3)I2, 7.0(3)I3, 7.0(3)I4, 7.0(3)I5, 7.0(3)I6`

shermdog: The tests are skipped entirely for I2-6.  Support for the feature was added in I7.

tests/beaker_tests/cisco_ip_multicast/test_ip_multicast.rb
```diff
@@ -37,6 +37,9 @@
 # Skip -ALL- tests if feature is not supported by image version
 skip_nexus_image(/I[2-6]/, tests)
```
shermdog: @mikewiebe this is where the test overall is skipped for I2-6

tests/beaker_tests/cisco_ip_multicast/test_ip_multicast.rb
mikewiebe: Ah, good point. Thanks for the reminder.

mikewiebe:
👍

PR #495, opened 2018-07-19
(maint) update test_package ordering

shermdog:
We have seen transient failures with these tests in CI
Set order of resources to ensure things are done in the correct order

```
Error: /Stage[main]/Main/Node[default]/Package[demo-one]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/yum -d 0 -e 0 -y list demo-one' returned 1: Error: No matching Packages to list
```


mikewiebe:
@shermdog 👍 Looks good.  I ran this through our CI (OAC, bash-shell, guestshell) just to be safe and the tests passed.

Issue #497, opened 2018-08-03
cisco_interface 100GB speed setting incorrect


gkevlin:
https://github.com/cisco/cisco-network-puppet-module/blob/develop/lib/puppet/type/cisco_interface.rb
 
Lines 183-187
```
 newproperty(:speed) do
    desc "Configure the speed between interfaces. Default value is 'auto'."

    newvalues(:auto, 10, 100, 1000, 10_000, **1_000_000**, 40_000, :default)
  end # property speed
```

1_000_000 is not a valid value.  For 100Gb interfaces the value should be 100_000.

If possible could this be fixed as part of the 1.9 branch with a 1.9.1 tag release?

Thanks,
George Kevlin

shermdog:
@gkevlin Thanks for the report.  We'll work on getting this fixed up.  I suspect it will be part of the upcoming 1.10.0 release.

cc @mikewiebe

mikewiebe:
@gkevlin @shermdog Is correct that we can include this as part of the `1.10` release that will be out in a few weeks.

mikewiebe:
Fixed in https://github.com/cisco/cisco-network-puppet-module/pull/499

PR #499, opened 2018-08-08
fix for issue cisco #497

saichint:
This PR is for fixing 
https://github.com/cisco/cisco-network-puppet-module/issues/497

The wrong speed problem exists in the network_interface as well. It is corrected too,

### DIFF
--- a/README.md
+++ b/README.md
@@ -2286,7 +2286,7 @@ cisco_interface { 'ethernet1/10':
 ```
 
 ###### `speed`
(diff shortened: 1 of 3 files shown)


saichint:
@mikewiebe No, I could not test it as our beaker tests don't have anything specific to 100G interfaces, and also our lab devices where we are running our tests, do not have 100G support cards as of now.
Since this is an enum change only, it should be easy to verify if the submitter has the setup.

@gkevlin Can you please take this PR and verify if this solves your issue?
Thanks

PR #502, opened 2018-08-23
(CISCO-75/76) Fix cisco_interface_hsrp_group

shermdog:
Removed default values for authentication. There are not defaults that are sane in this context. Users should supply explicit values.
Fixed authentication and ipv4/ipv6 validation

Requires cisco/cisco-network-node-utils#589


mikewiebe:
👍

PR #505, opened 2018-08-29
(maint) pin net-telnet version

shermdog:
The 0.2.0 release of net-telnet requires ruby >= 2.3.  We need to test ruby 1.9.3 until 31-Dec-18

Also bumped the travis rvm versions to match what is currently supported in puppet-agent.

@mikewiebe @willmeek this should resolve the travis issues that have popped up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,5 @@ language: ruby
 cache: bundler
 
 rvm:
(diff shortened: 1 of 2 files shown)


mikewiebe:
👍

PR #506, opened 2018-08-31
(CISCO-63) Confine package provider and fix inspector

shermdog:
Facter was updated to identify guestshell as nexus - confine the operating system so we do not try to inspect with cisco provider on Puppet Device Nodes
Also cleaned up error from package inspector poking around vs managing resources


mikewiebe:
@shermdog Will merge after conflicts are resolved.

shermdog:
@mikewiebe rebased.

PR #510, opened 2018-09-10
fix test_route_map beaker test for I9

rahushen:
Fixes the test_route_map ordering issue on 9k.

### DIFF
--- a/tests/beaker_tests/cisco_route_map/test_route_map.rb
+++ b/tests/beaker_tests/cisco_route_map/test_route_map.rb
@@ -26,7 +26,7 @@ require File.expand_path('../../lib/utilitylib.rb', __FILE__)
 
 # In I7 match_src_proto order is not maintained in running config.
 # This behavior is currently observed only on the N9K.
-if platform[/n9k/] && image_version.to_s.strip[/I7/]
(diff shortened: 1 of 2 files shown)


tests/beaker_tests/cisco_route_map/test_route_map.rb
```diff
@@ -26,7 +26,7 @@
 
 # In I7 match_src_proto order is not maintained in running config.
 # This behavior is currently observed only on the N9K.
-if platform[/n9k/] && image_version.to_s.strip[/I7/]
+if platform[/n9k/] && image_version.to_s.strip[/I(7|8|9)/]
```
saichint: Does this work for switches running 9.2.1 code? (One of my switch is running 9.2.1 and it fails there). There is no 'I9' in the show version for that switch.

rahushen: Yes, works for me with this image on:
NXOS: version 9.2(1)

saichint: Ok, In that case  👍

saichint:
👍

PR #517, opened 2018-10-12
Fixes for n9k-f and n3k-f

saichint:
* This PR is for fixing beaker tests and software versioning scheme which changed from older versions to latest versions on n3k-f and n9k-f.
* Some of the features like vtp, acl, vxlan_vtp_vni etc. have a different behavior for these platforms and this is addressed.

All tests pass on n9k, n9k-f, n3k-f

### DIFF
--- a/examples/cisco/demo_vtp.pp
+++ b/examples/cisco/demo_vtp.pp
@@ -16,7 +16,7 @@
 
 class ciscopuppet::cisco::demo_vtp {
 
-  if platform_get() =~ /n(3|5|6|7|8|9)k$/ {
+  if platform_get() =~ /n(3|5|6|7|9)k$/ {
 
     cisco_vtp { 'default':
       ensure   => present,
--- a/examples/cisco/demo_vtp.pp
+++ b/examples/cisco/demo_vtp.pp
@@ -15,11 +15,16 @@
 # limitations under the License.
 
 class ciscopuppet::cisco::demo_vtp {
-  cisco_vtp { 'default':
-    ensure   => present,
-    domain   => 'cisco1234',
-    password => 'test1234',
-    version  => 2,
-    filename => 'bootflash:/vlan.dat'
+
+  if platform_get() =~ /n(3|5|6|7|8|9)k$/ {
+
+    cisco_vtp { 'default':
+      ensure   => present,
+      domain   => 'cisco1234',
+      password => 'test1234',
+      version  => 2,
+      filename => 'bootflash:/vlan.dat'
+  } else {
+    notify{'SKIP: This platform does not support vtp': }
   }
 }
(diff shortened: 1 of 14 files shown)


examples/cisco/demo_vtp.pp
```diff
@@ -15,11 +15,16 @@
...
 
 class ciscopuppet::cisco::demo_vtp {
-  cisco_vtp { 'default':
-    ensure   => present,
-    domain   => 'cisco1234',
-    password => 'test1234',
-    version  => 2,
-    filename => 'bootflash:/vlan.dat'
+
+  if platform_get() =~ /n(3|5|6|7|8|9)k$/ {
```
mikewiebe: We should be able to remove 8.

lib/puppet/parser/functions/platform_fretta.rb
```diff
@@ -25,11 +25,13 @@ module Parser
     #
     module Functions
       newfunction(:platform_fretta, type: :rvalue) do |_args|
-        data = lookupvar('os')
+        data = lookupvar('cisco')
         return '' if data.nil?
 
-        pat = '7.0\(3\)F'
-        data['release']['full'].match(pat) ? true : false
+        data['inventory'].each do |_x, slot|
```
mikewiebe: Did you verify the demo manifest with this change?

lib/puppet/parser/functions/platform_fretta.rb
```diff
@@ -25,11 +25,13 @@ module Parser
...
+          return true if slot['pid'][/-R/]
```
mikewiebe: In the `node_utils` function we check for both `FM-R` and `-R`.  Do we need both checks here or can we simplify it to just check `-R`?

lib/puppet_x/cisco/cmnutils.rb
```diff
@@ -243,19 +243,27 @@ def self.range_summarize(range_str, sort=true)
         ranges.join(',').gsub('..', '-')
       end
 
+      # fretta check
+      def self.check_slot_pid(inv)
+        inv.each do |_x, slot|
+          return true if slot['pid'][/-R/]
```
mikewiebe: Same question here about `FM-R` vs `-R`

mikewiebe:
@saichint Just a few comments.

lib/puppet/parser/functions/platform_fretta.rb
saichint: Yes

lib/puppet/parser/functions/platform_fretta.rb
saichint: Changed NU to use -R only

lib/puppet_x/cisco/cmnutils.rb
saichint: Changed NU to use -R only

mikewiebe:
👍

Issue #518, opened 2018-10-17
resource type "cisco_interface" and vpc_id


henriots:
Hello!

When using resource type "cisco_interface" for setting VPC IDs for port-channels it fails every time after first run, because it doesn't check whether the ID has previously been set or not.

Error:
```
Could not evaluate: [interface port-channel 11] The command ' vpc 11' was rejected with error:
ERROR: Operation failed: [vPC already exists]
```

saichint:
@henriots Thanks for bringing this to our attention. It is possible that this is a platform specific issue. When I tested with newer platforms I don't see any issue. 
Can you try this fix?
Change this line:
https://github.com/cisco/cisco-network-puppet-module/blob/develop/lib/puppet/type/cisco_interface.rb#L284
from value to value.to_i

basically you can go to your puppet server and look for this file at:
/etc/puppetlabs/code/environments/production/modules/ciscopuppet/lib/puppet/type/cisco_interface.rb and do the change there and try it or change it in the source and reinstall and try it.
Let us know if this works.

saichint:
@henriots Instead of the above suggestion, please try this code in the PR I just opened:
https://github.com/cisco/cisco-network-puppet-module/pull/519
The reason is that vpc_id could be 'default' and it cannot be changed to integer.

henriots:
@saichint 
Hey!

Still the same.
First run:
```
Notice: /File[/opt/puppetlabs/puppet/cache/lib/puppet/type/cisco_interface.rb]/c                                                                                                                                                                                                                                             ontent:
--- /opt/puppetlabs/puppet/cache/lib/puppet/type/cisco_interface.rb     2018-10-                                                                                                                                                                                                                                             05 11:40:52.000000000 +0300
+++ /tmp/puppet-file20181018-14927-8wvy0j       2018-10-18 09:22:21.398815433 +0                                                                                                                                                                                                                                             300
@@ -277,9 +277,10 @@
     munge do |value|
       value = :default if value == 'default'
       if value != :default
+        value = value.to_i
         range = *(1..4096)
         fail 'VPC ID must be in the range 1..4096' unless
-          range.include?(value.to_i)
+          range.include?(value)
       end
       value
     end

[...]

PR #519, opened 2018-10-17
[WIP] fix vpc id for older platforms

saichint:
This PR is for fixing vpc_id idempotent issue for older platforms like n5k.


saichint:
This is not needed as #518 is a manifest problem.

PR #536, opened 2019-02-08
(maint) clearing down the node environment on new device instances

Thomas-Franklin:
Refer to https://github.com/cisco/cisco-network-node-utils/pull/601#issuecomment-440338013 for the explanation. This basically allows for multiple devices to be managed in the same catalog compilation without environmental bleeding.


Thomas-Franklin:
@mikewiebe this is just a tidy up for environment handling during `puppet device` runs if there is more than one device to manage in same compilation. I am unable to add you as a reviewer as I probably don't have contributor permissions.

PR #543, opened 2019-02-12
beaker: Add NVE probes to vrf/vlan tests; test_manifest hidden errors

chrisvanheuveln:
Mostly n6k test errors. Also noticed that `test_vrf` was not raising an error in `test_manifest` when it hit the `NVE Feature NOT supported on this Platform` failure, which passed because the `'Applied catalog'` string shows up in the output. The code logic also needs to know how to handle intended errors (negative testing) so my change is to explicitly look for `Error: ` in the output but only when the caller does not supply a `stderr_pattern`.

Tested with agentless on N6k; will update after testing with agentful.

### DIFF
--- a/tests/beaker_tests/lib/utilitylib.rb
+++ b/tests/beaker_tests/lib/utilitylib.rb
@@ -360,9 +360,10 @@ DEVICE
           #  raise 'Errored test'
           # end
           on(tests[:proxy_agent], puppet_device_cmd, acceptable_exit_codes: [0, 1, 2])
-          if !(stdout.include? 'Applied catalog') && tests[id][:stderr_pattern].nil?
+          if tests[id][:stderr_pattern].nil? && (output[/Error: /] || !output[/Applied catalog/])
+            logger.info(tests[id][:manifest])
             logger.info(stdout)
-            raise 'Errored test as the command result did not match applied catalog'
+            raise 'Unexpected error while applying catalog'
           end
         else
           on(tests[:agent], puppet_agent_cmd, acceptable_exit_codes: code)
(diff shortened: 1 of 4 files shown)


tests/beaker_tests/lib/utilitylib.rb
```diff
@@ -360,9 +360,10 @@ def test_manifest(tests, id)
           #  raise 'Errored test'
           # end
           on(tests[:proxy_agent], puppet_device_cmd, acceptable_exit_codes: [0, 1, 2])
-          if !(stdout.include? 'Applied catalog') && tests[id][:stderr_pattern].nil?
+          if tests[id][:stderr_pattern].nil? && (output[/Error: /] || !output[/Applied catalog/])
```
mikewiebe: This does not look right to me.  When using the `on` agentfull method you should use `stdout` instead of `output` right?

chrisvanheuveln: Good catch. It's a cut&paste from the agentless codepath. I'm still setting up the agentfull TB so I'll fix & confirm there.

chrisvanheuveln:
Successful agent-full run:
https://earms-trade.cisco.com/tradeui/logs/logfile?logfile=TaskLog.AgentsTests&begin=98718&size=104844&archive=%2Fauto%2Fatslogs%2Fcl%2F2019%2F02%2F12%2F17%2F02%2Fagents_job.2019Feb12_17:02:06.064636.zip&ats=%2Fauto%2Frtp-core%2Fats_shared%2Fusers%2Fcvanheuv&submitter=cvanheuv

PR #551, opened 2019-05-22
Transport class update work

willmeek:


willmeek:
See https://github.com/cisco/cisco-network-puppet-module/pull/550 for main comment thread

This has been created to push a function to allow for pre-transports credentials to be maintained

(No push access to @david22swan repo and on PTO)

willmeek:
cc @DavidS as the Transports expert :)

willmeek:
550 has been updated with the commit

PR #553, opened 2019-06-06
(maint) Minor fixes

willmeek:
Transport schema Port should be an Integer rather than a String.
The metadata refers to the Resource API gem rather than module and so minimum should be the 1.0.0 release


willmeek:
@mikewiebe @chrisvanheuveln minor fixes
cc @shermdog

PR #555, opened 2019-06-21
(maint) Remove extra require from Rakefile

willmeek:
PDK sync.yml updates the Rakefile to add an extra non-guarded puppet-strings/tasks.
Remove this.


willmeek:
@mikewiebe This should address the CI issue

mikewiebe:
@willmeek Thanks for the quick fix.

Issue #496, opened 2018-07-20
Very slow execution when managing interfaces


bsauvajon:
Hi,
I've started to use the ciscopuppet module to manage configuration on a nexus 5672UP with 12 FEX 2232TM-E, using Open Agent Container.
Firstly, I've successfully managed local users, and now I'm starting to manage some interfaces configuration.

As soon as I declared my first interface, manifest execution time has gone from 30 seconds to 900 seconds.
Debug shows puppet takes a very long time to scan and build the list of existing interface ressources on the switch. Each interface discovery seems to take between 1 and 2 seconds, and when adding physical interfaces, vlan interfaces an port-channel, I have more than 500 interfaces declared. Is there a way to accelerate this step ?

mikewiebe:
@bsauvajon Thanks for your interest in the module.  We will look into this problem to see if we can improve the performance.  Would you be willing to share your manifest and debug logs?

bsauvajon:
@mikewiebe No problem to share manifest and debug logs, how can I send it ?

mikewiebe:
@bsauvajon Let me try and recreate it in my environment first with the information you have already provided and if I cannot then I will reach out to you for more information.

mikewiebe:
@bsauvajon Just an update on this issue.  I have been able to recreate the problem and am investigating potential solutions.

mikewiebe:
Resolved by https://github.com/cisco/cisco-network-puppet-module/pull/561

PR #557, opened 2019-07-12
cisco_interface: interface lookup performance enhancement

chrisvanheuveln:
#### Summary
- The existing behavior in `cisco_interface` provider code's `prefetch` method is to collect state for all interfaces on the device and to cache the data for other interface lookups. This approach makes sense in many cases but may result in unnecessarily long processing time for use cases where there are thousands of interfaces while the manifest is only interested in a small subset of interfaces. (Ref: https://github.com/cisco/cisco-network-puppet-module/issues/496)
- This PR creates a compromise by adding a per-interface lookup path to `prefetch` that is invoked when the manifest's `cisco_interface` resource count is 10 or less. This value is arbitrary -- it would be ideal to specify this as a top-level resources value in the manifest that all providers could consume, but there doesn't appear to be a vehicle for that yet. I've reached out to Puppetlabs for possible solutions.
- Supporting changes have already been made to the node_utilities interface code. (Ref: https://github.com/cisco/cisco-network-node-utils/pull/630)
- Beaker test `test_interface_threshold.rb` was created to verify that the two code-paths are taken when under and over the interface count threshold. This will be leveraged for related changes (in progress) for other interface providers that need this performance change.
- There were a number of interface-related deprecated properties remaining in the `cisco_interface` provider. These have been removed.


lib/puppet/provider/cisco_interface/cisco.rb
```diff
@@ -177,16 +143,31 @@
 
   def initialize(value={})
     super(value)
-    @nu = Cisco::Interface.interfaces[@property_hash[:name]]
+    if value.is_a?(Hash)
+      # value is_a hash when initialized from properties_get()
+      all_intf = value[:all_intf]
+      single_intf = value[:interface]
+    else
+      # note: @nu instantiation will be skipped
```
mikewiebe: Why is @nu skipped in this case?

lib/puppet/provider/cisco_interface/cisco.rb
```diff
@@ -203,24 +184,36 @@ def self.properties_get(interface_name, nu_obj)
...
-      resources[name].provider = provider unless provider.nil?
+    if resources.keys.length > 10
+      interfaces = instances
+      info "[prefetch all-interfaces: #{interfaces.length}]"
+      resources.keys.each do |name|
+        provider = interfaces.find { |intf| intf.instance_name == name }
+        resources[name].provider = provider unless provider.nil?
+      end
+    else
+      info '[prefetch per-interface]'
```
mikewiebe: This `info` statement confuses me.  I assume this means prefetch one interface at a time?

lib/puppet/provider/cisco_interface/cisco.rb
mikewiebe: Clarify comment here.  What is really happening?

lib/puppet/provider/cisco_interface/cisco.rb
chrisvanheuveln: fixed

lib/puppet/provider/cisco_interface/cisco.rb
chrisvanheuveln: I'll replace the comment with this:
```
      # @property_hash[:name] is nil in this codepath; because it's nil
      # it will cause @nu to become nil, thus @nu instantiation is just
      # skipped altogether.
```

chrisvanheuveln:
Closing this PR because it's not picking up my latest commits (github suggests a problem with git-hooks), plus I'm adding some auto-threshold code which will remove the `@show_run_int_threshold` approach I was using earlier.

PR #559, opened 2019-07-18
(FM-8271) - Exceptions added to the code for Non-Default VDC's

david22swan:
- Test's that run the command `show run pdk` now skipped


chrisvanheuveln:
👍 
Sorry for the delay in reviewing this. Thanks for validating the non-vdc environment! This is very helpful.

PR #569, opened 2019-08-02
 move threshold checks into facter

chrisvanheuveln:
This PR includes fixes to support the `N7k` workaround. I also moved the threshold checking code into facter to simplify the provider files.

I haven't fully tested this code yet, just manual one-offs.


chrisvanheuveln:
I need to make a few more changes to this design.

PR #571, opened 2019-08-05
Fix snmp_user test for older images

mikewiebe:
This fixes a problem on older `/7.0.3.I2|I3|I4/` images.

A few image versions have a bug where the device displays the previously configured auth protocol.  This fix removes the snmp user before executing non-default property test for the impacted versions.


mikewiebe:
@chrisvanheuveln Gave me verbal approval to commit this.

PR #572, opened 2019-08-06
Move threshold logic into cmnutils

chrisvanheuveln:
Tested on n9k-108/GS and n6k-77/Agtless.

### DIFF
--- a/lib/facter/cisco_nexus.rb
+++ b/lib/facter/cisco_nexus.rb
@@ -39,23 +39,14 @@ class Facter::CiscoNexus
     # getting all interfaces at once. The threshold is only useful to a certain
     # point - it depends on the total number of interfaces on the device - after
     # which it's better to just get all interfaces.
-    facts['interface_count'] = count = Cisco::Interface.interface_count
-    if Gem::Version.new(CiscoNodeUtils::VERSION) > Gem::Version.new('2.0.2')
-      facts['single_intf_support'] = true
-      if count < 1000 && facts['hardware']['type'][/Nexus7/]
-        # N7 uses a less efficient show cmd get_value to workaround an image bug
-        thresh_pct = 0.075
-      else
-        thresh_pct = 0.15
-      end
+    facts['interface_count'] = Cisco::Interface.interface_count
+    if facts['interface_count'] < 1000 && facts['hardware']['type'][/Nexus7/]
+      # N7 uses a less efficient show cmd get_value to workaround an image bug
+      thresh_pct = 0.075
     else
-      facts['single_intf_support'] = false
-      thresh_pct = 0
-      info '## Notice: Unable to prefetch interfaces independently, using legacy lookup instead.'
-      info '## Notice: cisco_node_utils gem does not contain interface lookup enhancements.'
-      info '## Notice: Please upgrade cisco_node_utils to v2.1.0 or newer'
+      thresh_pct = 0.15
     end
-    facts['show_run_int_threshold'] = count * thresh_pct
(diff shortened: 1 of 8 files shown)


chrisvanheuveln:
Hold off on review... WIP.

chrisvanheuveln:
okay, review it!

lib/puppet/provider/cisco_interface/cisco.rb
```diff
@@ -185,19 +186,17 @@ def self.properties_get(interface_name, nu_obj, all_intf: nil)
     new(current_state)
   end # self.properties_get
 
-  def self.instances(single_intf=nil)
+  def self.instances(single_intf=nil, interface_threshold=0)
```
mikewiebe: Do you really need the default arg value for `interface_threshold` since you have `facts['interface_threshold'] = 0` in `self.interface_threshold`?

lib/puppet/provider/cisco_interface/cisco.rb
```diff
@@ -185,19 +186,17 @@ def self.properties_get(interface_name, nu_obj, all_intf: nil)
...
     # 'puppet resource' calls here directly; will always get all interfaces.
     # 'puppet agent' callpath is initialize->prefetch; may pass a single intf.
-    all_intf = single_intf ? false : true
-    interfaces = []
-    if Gem::Version.new(CiscoNodeUtils::VERSION) > Gem::Version.new('2.0.2')
+    if single_intf && interface_threshold > 0
+      all_intf = false
       nu_interfaces = Cisco::Interface.interfaces(nil, single_intf)
     else
-      info '## Notice: Unable to prefetch independently, using legacy lookup instead.'
```
mikewiebe: Why did these first two notices not make it into the method above?  I feel like these add good information and would prefer we keep them.

lib/facter/cisco_nexus.rb
```diff
@@ -50,4 +52,27 @@ def self.platform_facts
...
+
+  def self.interface_threshold(hw_type)
+    # Count interfaces and compute a lookup efficiency threshold. The threshold
+    # is a cutoff for determining when to get each interface one at a time versus
+    # getting all interfaces at once. The threshold is only useful to a certain
+    # point - it depends on the total number of interfaces on the device - after
+    # which it's better to just get all interfaces.
+    facts = {}
+    if Gem::Version.new(CiscoNodeUtils::VERSION) <= Gem::Version.new('2.0.2')
+      info '## Notice: Please upgrade cisco_node_utils to v2.1.0 or newer'
```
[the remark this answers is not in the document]
chrisvanheuveln: You mean like this?

```
info '## Notice: Please upgrade cisco_node_utils gem to v2.1.0 or newer'
                                                 ^^^
```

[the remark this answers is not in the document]
chrisvanheuveln: fxied

lib/puppet/provider/cisco_interface/cisco.rb
chrisvanheuveln: Yes and no. `instances` may be called directly via `puppet resource` so it's best to initialize it here (gotta protect my api's ya know), but the way the logic works `interface_threshold` should only be referenced when `single_intf` is also defined. So really it's a best practice to initialize it here.

lib/puppet/provider/cisco_interface/cisco.rb
chrisvanheuveln: They live in `cmnutils` now. I moved them there so that I wouldn't need to specify them in each provider.

PR #581, opened 2019-10-17
(MAINT) Update ordering of transport parameters

npwalker:
The ordering of the parameters here is used to determine the ordering of the parameters when displayed in the PE Console UI.  Since user and password are required and likely more utilized than transport and verify mode we move them up.


DavidS (approved):
:+1:

